Issue #15, opened 2024-10-04
Sorting of signatures


ryanrolds:
Sort both list of signatures by commonly (Wanderer and in-game) aids cognition. It would be useful if the sigantures where sorted by ID by default and there were column headers that allowed changing the sort.

I'm an experienced software engineer, but have limited experience with Elixir/Phoenix. If you're willing to assume the burden of telling me what I'm doing wrong, I'm happy to work on this.

DmitryPopov:
Signature list sorting can be implemented on ui widget (react.js) level actually, we have all signatures data here.

ryanrolds:
I see. I haven't looked at PrimeReact before. Nifty componant libriary. I will have a PR up that adds sortable columns to the DataTable.

ryanrolds:
Do you have any objections to the default sort being ID and not age?

DmitryPopov:
You can add a property to set the default sorting, so we can decide it later.

DanSylvest:
Hi! That's cool, but my suggestion give ability to sort by dropdown menu in header of widget (yes that's dropdown will be very small) or add it in settings of signatures. Now functional of signatures not finished and I working on it now.

ryanrolds:
> You can add a property to set the default sorting, so we can decide later.

Will so. Should the current sort be perisisted between sesions?

> Hi! That's cool, but my suggestion give ability to sort by dropdown menu in header of widget (yes that's dropdown will be very small) or add it in settings of signatures. Now functional of signatures not finished and I working on it now.

I'm hearing that the the goal to keep the UI small to ensure many signatures are visiable. A dropdown on the right of the title bar makes sense too and saves about a row and half of signtuare space. I assume button to toggle lazy delete would go there as well. How would reverse be implemented in a dropdown? Additional entries?

DmitryPopov:
> Will so. Should the current sort be perisisted between sesions?

I guess so, better to keep user settings as we do for other user-related settings (using LocalStorage).

PR #18, opened 2024-10-05
Sortable signatures

ryanrolds:
Re: https://github.com/wanderer-industries/wanderer/issues/15

Changes:
* Show header
* Marked columns as sortable and set custom style to reduce header size (sadly setting DataTable's size doesn't also control the header)

Oddities:
* Reversing default sort doesn't do anything until sorted by another column first. This may be a bug in PrimeReact.


ryanrolds:
I'm going to work on controlling the default in another PR.

DanSylvest:
Hi! Headers and sigs have same size, yes?

DanSylvest (approved):
it looks good

PR #20, opened 2024-10-06
Persist the signature sort between sessions and header improvements

ryanrolds:
* Stores signature field and direction in local storage
* Sets initial sort using value in local storage (or default)
* Setting the sort order fixed the issue with reversing default sort not working until selecting another column
* Fixed header alignment
* Improved the header display a little
* Moved the header CSS to fixes.scss.


DanSylvest:
Hi, @ryanrolds i've merged your PR. So, good job! 👍 

In next PR i just little bit fix styles. :)

PR #87, opened 2024-12-31
Allow user to follow a specific tracked character

guarzo:
This PR adds functionality to allow users to choose a single character to be "followed" and have the systems they enter, newly splashed or not, selected in the UI.   This is useful for multiboxing situations where you may be moving around multiple characters, but only want to paste signatures with your scanner.     

I wasn't sure on the best UI, so I left a previous commit in place that shows another option

UI in pull request (highlight to follow)
highlight

Other option (radio buttons for follow)

radio


DmitryPopov:
I'd prefer more explicit options to have radio button selecting follow char.

lib/wanderer_app/character/tracker.ex
```diff
@@ -526,6 +528,14 @@ defmodule WandererApp.Character.Tracker do
 
         WandererApp.Character.update_character(character_id, location)
 
+        for map_id <- active_maps do {
+          MapServer.broadcast!(map_id, :maybe_select_system, %{
```
DmitryPopov, line 532: Character tracker implementation shouldn't know about/be linked to MapServer implementation, so please consider moving selection logic to Map.Server.ConnectionsImpl (on update location logic) - you can check if character is marked as following here as well

DmitryPopov (changes requested):

lib/wanderer_app/map/server/map_server_characters_impl.ex
```diff
@@ -403,6 +404,11 @@ defmodule WandererApp.Map.Server.CharactersImpl do
           [
             {:character_location, %{solar_system_id: solar_system_id}, %{solar_system_id: nil}}
           ]
+          # broadcast for all location changes, but will only select system for characters that are followed
```
guarzo, line 407: this seems cleaner than checking followed status here, and it needed to be CharactersImpl rather than ConnectionsImpl as connections is only valid for a new connection, rather than any location change

DmitryPopov: Agreed, but better to check it inside true block:
```
WandererApp.Cache.lookup!(
      "character:#{character_id}:location_started",
      false
    )
    |> case do
      true ->
              <here>
       
      false ->
              <not here (you suggested)>
```

lib/wanderer_app/character/tracker.ex
```diff
@@ -682,7 +682,7 @@ defmodule WandererApp.Character.Tracker do
 
   defp maybe_update_active_maps(
          %{character_id: character_id, active_maps: active_maps} = state,
-         %{map_id: map_id, track: false} = _track_settings
+         %{map_id: map_id, track: false, followed: false} = _track_settings
```
DmitryPopov, line 685: I guess u don't need to match for 'followed: false' here, it can be removed.

lib/wanderer_app/character/tracker_manager_impl.ex
```diff
@@ -420,7 +420,8 @@ defmodule WandererApp.Character.TrackerManager.Impl do
         {:ok, character_state} =
           WandererApp.Character.Tracker.update_track_settings(character_id, %{
             map_id: map_id,
-            track: false
+            track: false,
+            followed: false
```
DmitryPopov, line 424: Same here, 'followed: false' can be removed

lib/wanderer_app/map/server/map_server_characters_impl.ex
```diff
@@ -78,7 +77,8 @@ defmodule WandererApp.Map.Server.CharactersImpl do
       |> Enum.each(fn character_id ->
         WandererApp.Character.TrackerManager.update_track_settings(character_id, %{
           map_id: map_id,
-          track: false
+          track: false,
+          followed: false
```
DmitryPopov, line 81: Same here, 'followed: false' can be removed

lib/wanderer_app/map/server/map_server_characters_impl.ex
```diff
@@ -297,6 +297,7 @@ defmodule WandererApp.Map.Server.CharactersImpl do
       WandererApp.Character.TrackerManager.update_track_settings(character_id, %{
         map_id: map_id,
         track: true,
+        followed: false,
```
DmitryPopov, line 300: Same here, 'followed: false' can be removed

DmitryPopov (changes requested):

PR #95, opened 2025-01-04
fix(map): prevent deselect on click to map

### LINKED ISSUE #80: Don't deselect system on click to map

guarzo:
Fixes #80
Fixes: #91

- Prevents single node deselection on background / same node click
- Allows deselection of all nodes if multiple are currently selected

### DIFF
--- a/assets/js/hooks/Mapper/components/map/Map.tsx
+++ b/assets/js/hooks/Mapper/components/map/Map.tsx
@@ -119,7 +119,7 @@ const MapComp = ({
   isSoftBackground,
   onAddSystem,
 }: MapCompProps) => {
-  const { getNode } = useReactFlow();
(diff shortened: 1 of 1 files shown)


assets/js/hooks/Mapper/components/map/Map.tsx
```diff
@@ -186,6 +186,12 @@ const MapComp = ({
     (changes: NodeChange[]) => {
       const systemsIdsToRemove: string[] = [];
 
+      // prevents single node deselection on background / same node click
+      // allows deseletion of all nodes if multiple are currently selected
+      if (changes.length === 1 && changes[0].type == 'select' && changes[0].selected === false) {
```
guarzo, line 191: could instead prevent deselection at all on background click, but I think this is cleaner, and it's useful to be able to deselect multiple items quickly

PR #97, opened 2025-01-05
 feat(map):  Enables left click select and right click pan and drag

guarzo:
Adds user interface setting to toggle using the left mouse drag for selection and right click drag for pan.  Also adjusts css to have the selection box disappear after drag stop as nodes remain selected with highlighted boundary boxes.

PanAndDrag
GifDemo


guarzo:
better to control this via theme

PR #100, opened 2025-01-07
feat: add theme selection, and pathfinder theme

guarzo:


assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNode.tsx
```diff
@@ -193,23 +193,33 @@ export const SolarSystemNode = memo(({ data, selected }: WrapNodeProps<MapSolarS
...
       <div
-        className={clsx(classes.RootCustomNode, regionClass, classes[STATUS_CLASSES[status]], {
-          [classes.selected]: selected,
-        })}
+        className={clsx(
+          classes.RootCustomNode,
+          regionClass,
+          classes[STATUS_CLASSES[status]],
+          { [classes.selected]: selected },
+        )}
```
DanSylvest, line 201: I think we should still to use tailwind classes instead creating a lot of styles in *.scss file. Just thing for think.
Any classes of tailwind you may find here: https://tailwindcss.com/docs/installation

It's regarding all similar cases neither.

DanSylvest (approved):
Anyway. Great work, let's add this to prod

DanSylvest (approved):

PR #107, opened 2025-01-09
fix: fix route color

guarzo:
- also updates dark mode in default theme, and some pathfinder theme settings


assets/js/hooks/Mapper/components/mapInterface/widgets/RoutesWidget/RoutesList/RoutesList.module.scss
```diff
@@ -8,7 +8,6 @@
 .RouteSystem {
   width: 8px;
   height: 8px;
-  background: #ffffff;
```
guarzo, line 11: this was overriding the route color

PR #113, opened 2025-01-10
feat(widget): add structure widget #46

guarzo:
structure-edit

structure-with-timer

- added api for system timers as well


assets/js/hooks/Mapper/components/mapInterface/widgets/SystemStructures/helpers/structureUtils.ts
```diff
@@ -0,0 +1,60 @@
...
+      removed.push(oldItem);
+    }
+  }
+
+  return { added, updated, removed };
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function mapServerStructure(serverData: any): StructureItem {
+  console.debug(`server data: ${JSON.stringify(serverData)}`);
```
DmitryPopov, line 31: Better to get rid of debug/console logs.

DmitryPopov (changes requested):

guarzo:
> Would love this but the app is already fairly widget heavy so an option to hide this or minimize it would be helpful. ❤️ edit: pulled that minimize idea out of no where but now that i think about it i would love that on all the widgets so i can maximize my screen space when possible

disabled

assets/js/hooks/Mapper/components/mapInterface/MapInterface.tsx
```diff
@@ -1,17 +1,22 @@
...
 import { CURRENT_WINDOWS_VERSION, DEFAULT_WIDGETS } from '@/hooks/Mapper/components/mapInterface/constants.tsx';
 import { useMapRootState } from '@/hooks/Mapper/mapRootProvider';
+import { WidgetsIds } from '@/hooks/Mapper/components/mapInterface/constants.tsx';
 
 type WindowsLS = {
   windows: WindowProps[];
   version: number;
 };
 
+export interface MapInterfaceHandle {
```
guarzo, line 16: @DanSylvest  @DmitryPopov  -- after the window change -- seems like only default widgets will show regardless of the setting, this works, but maybe a better approach is possible?

PR #117, opened 2025-01-11
Clean-up theme swap logic for nodes

guarzo:


guarzo:
doing some more testing

guarzo:
more testing

PR #126, opened 2025-01-17
feat: add local pilots display on hover

guarzo:
image

- displays a tooltip with tracked pilots in system, including ship type and ship name


guarzo:
merged this with zkilltooltip PR to solve for the merge conflcts

PR #128, opened 2025-01-17
Hover tooltip for kills bookmark and local counter

guarzo:
image

image

-- adds tooltip on kills bookmark
-- displays a tooltip with tracked pilots in system, including ship type and ship name


guarzo:
swapped to cleaner branch

PR #194, opened 2025-02-20
feat: Auto-set connection EOL status and ship size when linking/editing signatures

alpha02x:
Automatically updates wormhole connections both in the signature linking dialog and when manually editing with signature settings

### DIFF
--- a/assets/js/hooks/Mapper/helpers/getWhSize.ts
+++ b/assets/js/hooks/Mapper/helpers/getWhSize.ts
@@ -1,13 +1,6 @@
+import { SHIP_MASSES_SIZE } from '../components/map/constants';
 import { ShipSizeStatus } from '../types/connection';
 import { WormholeDataRaw } from '../types/wormholes';
-
-const SIZE_CLASSES = [
-  { threshold: 5_000_000, status: ShipSizeStatus.small },
-  { threshold: 62_000_000, status: ShipSizeStatus.medium },
-  { threshold: 375_000_000, status: ShipSizeStatus.large },
-  { threshold: 1_000_000_000, status: ShipSizeStatus.freight },
-  { threshold: 2_000_000_000, status: ShipSizeStatus.capital },
-];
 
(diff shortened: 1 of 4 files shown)


DmitryPopov:
We'll review that one, as far as getting a stable version.

DmitryPopov:
@DanSylvest please take a look

assets/js/hooks/Mapper/helpers/getWhSize.ts
```diff
@@ -0,0 +1,26 @@
...
+import { WormholeDataRaw } from '../types/wormholes';
+
+const SIZE_CLASSES = [
+  { threshold: 5_000_000, status: ShipSizeStatus.small },
+  { threshold: 62_000_000, status: ShipSizeStatus.medium },
+  { threshold: 375_000_000, status: ShipSizeStatus.large },
+  { threshold: 1_000_000_000, status: ShipSizeStatus.freight },
+  { threshold: 2_000_000_000, status: ShipSizeStatus.capital },
+];
+
```
DanSylvest, line 11: Hi! 
First. Please describe this constants here:
`assets/js/hooks/Mapper/components/map/constants.ts`

and check comment next

assets/js/hooks/Mapper/helpers/getWhSize.ts
```diff
@@ -0,0 +1,26 @@
...
+export const getWhSize = (whDatas: WormholeDataRaw[], whType: string): ShipSizeStatus | null => {
+  if (whType === 'K162' || whType == null) return null;
+
+  const wormholeData = whDatas.find(wh => wh.name === whType);
+
+  if (!wormholeData?.max_mass_per_jump) return null;
+
+  for (const sizeClass of SIZE_CLASSES) {
```
DanSylvest, line 19: Secnd. I didn't hear about another sizes of wh, i how I think we should not using math comparing operators. 
Better make like this.
```
const SIZE_CLASSES = {
  5000000: ShipSizeStatus.small,
  //  ...
};

// and use it like that
const size = SIZE_CLASSES[sizeClass.threshold];
```

PR #195, opened 2025-02-20
fix: prevent constant full signature widget rerender

guarzo:
- current behavior matches original,, still have re-render for the time change, system, lazy delete

https://github.com/user-attachments/assets/53d1769f-492b-40bd-bcfd-90ea03bf35ec

### DIFF
--- a/assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/SystemSignatures.tsx
+++ b/assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/SystemSignatures.tsx
@@ -85,7 +85,7 @@ export const SystemSignatures: React.FC = () => {
 
   const [sigCount, setSigCount] = useState<number>(0);
   const [pendingSigs, setPendingSigs] = useState<SystemSignature[]>([]);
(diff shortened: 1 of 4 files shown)


assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/SystemSignatures.tsx
```diff
@@ -85,7 +85,7 @@ export const SystemSignatures: React.FC = () => {
 
   const [sigCount, setSigCount] = useState<number>(0);
   const [pendingSigs, setPendingSigs] = useState<SystemSignature[]>([]);
-  const [undoPending, setUndoPending] = useState<() => void>(() => () => {});
+  const [undoPendingFn, setUndoPendingFn] = useState<() => void>(() => () => {});
```
DmitryPopov, line 88: Better to useRef for functions here

guarzo: going to memo the full header, but good calll

PR #196, opened 2025-02-20
Prevent unexpected system removal

guarzo:


assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/hooks/useSystemSignaturesData.ts
```diff
@@ -159,21 +159,6 @@ export function useSystemSignaturesData({
     onPendingChange?.(combined, undoPending);
   }, [localPendingDeletions, pendingUndoAdditions, onPendingChange, undoPending]);
 
-  useEffect(() => {
```
guarzo, line 162: only recent change that might be creating some sort of race condition with system_cleanup and splash?

lib/wanderer_app_web/live/maps/event_handlers/map_systems_event_handler.ex
```diff
@@ -58,6 +58,17 @@ defmodule WandererAppWeb.MapSystemsEventHandler do
       socket
     else
       # Check if we already selected this exact system for this char:
+      case WandererApp.MapSystemRepo.get_by_map_and_solar_system_id(map_id, solar_system_id) do
```
guarzo, line 61: force update last activity on splash?

PR #197, opened 2025-02-20
Add back signature expiration

guarzo:
- ensures system activity is updated on splash, and adds a delay on the signature deletion


DmitryPopov:
I'd implemented all sigs expiration logic on server side, same way as connections expiration logic works.

PR #262, opened 2025-03-16
feat: add static wh info

guarzo:


lib/wanderer_app_web/controllers/common_api_controller.ex
```diff
@@ -5,6 +5,13 @@ defmodule WandererAppWeb.CommonAPIController do
   alias WandererApp.CachedInfo
   alias WandererAppWeb.UtilAPIController, as: Util
 
+  # Cache wormhole data at module compilation time
+  @wormhole_types WandererApp.EveDataService.load_wormhole_types()
```
guarzo, line 9: @DmitryPopov  is this ok?

lib/wanderer_app_web/controllers/common_api_controller.ex
```diff
@@ -4,6 +4,14 @@ defmodule WandererAppWeb.CommonAPIController do
 
   alias WandererApp.CachedInfo
   alias WandererAppWeb.UtilAPIController, as: Util
+  alias WandererApp.EveDataService
+
+  # Cache wormhole data at module compilation time
+  @wormhole_types WandererApp.EveDataService.load_wormhole_types()
```
DmitryPopov, line 10: use WandererApp.CachedInfo.get_wormhole_types instead

guarzo: done

PR #288, opened 2025-03-19
fix: cloak key error behavior

guarzo:


lib/wanderer_app/vault.ex
```diff
@@ -3,20 +3,25 @@ defmodule WandererApp.Vault do
...
+          iv_length: 12
         }
       )
 
     {:ok, config}
   end
 
   defp decode_env!(var) do
-    var
-    |> System.get_env("OtPJXGfKNyOMWI7TdpcWgOlyNtD9AGSfoAdvEuTQIno=")
```
guarzo, line 19: pretty sure this isn't the intended name of the env var

PR #301, opened 2025-03-21
feat (api) add map connections endpoint

tysongg:


lib/wanderer_app_web/controllers/map_api_controller.ex
```diff
@@ -61,6 +62,44 @@ defmodule WandererAppWeb.MapAPIController do
...
+    properties: %{
+      id: %OpenApiSpex.Schema{type: :string},
+      map_id: %OpenApiSpex.Schema{type: :string},
+      solar_system_source: %OpenApiSpex.Schema{type: :integer},
+      solar_system_target: %OpenApiSpex.Schema{type: :integer},
+      mass_status: %OpenApiSpex.Schema{type: :integer},
+      time_status: %OpenApiSpex.Schema{type: :integer},
+      ship_size_type: %OpenApiSpex.Schema{type: :integer},
+      type: %OpenApiSpex.Schema{type: :integer},
+      womrhole_type: %OpenApiSpex.Schema{type: :string},
```
DmitryPopov, line 77: typo here: womrhole_type -> wormhole_type

DmitryPopov (changes requested):

lib/wanderer_app_web/controllers/map_api_controller.ex
tysongg: Good catch, fixed.

PR #323, opened 2025-04-14
Add temp name to signature and pass it to linked system

alpha02x:


assets/js/hooks/Mapper/components/mapRootContent/components/SignatureSettings/components/SignatureGroupContentWormholes.tsx
```diff
@@ -32,6 +35,19 @@ export const SignatureGroupContentWormholes = () => {
         <span>EOL:</span>
         <SignatureEOLCheckbox name="isEOL" />
       </label>
+
+      {isTempSystemNameEnabled && (
```
DanSylvest, line 39: This part of code should be moved into separate Component. As it was made for another elements.

alpha02x: ✓

DmitryPopov:
@alpha02x we're changing our flow to add new features a bit. Please create an issue with quick description/cases on our GH project first, so we can discuss what options will be available by default or can be added later.

alpha02x:
Here goes the issue: https://github.com/wanderer-industries/wanderer/issues/339
My discord name is the same as my github name

PR #351, opened 2025-05-01
feat (api): add additional system/connections methods

guarzo:


lib/wanderer_app_web/controllers/map_api_controller.ex
guarzo, line 1: split map controller due to size

lib/wanderer_app_web/controllers/map_system_api_controller.ex
```diff
@@ -0,0 +1,298 @@
...
+    parameters: [map_slug: [in: :path], map_id: [in: :path], id: [in: :path]],
+    responses: ResponseSchemas.standard_responses(@detail_response_schema)
+  def show(%{assigns: %{map_id: map_id}} = conn, %{"id" => id}) do
+    with {:ok, system_id} <- APIUtils.parse_int(id),
+         {:ok, system} <- Operations.get_system(map_id, system_id) do
+      APIUtils.respond_data(conn, APIUtils.map_system_to_json(system))
+    end
+  end
+
+  operation :create,
```
guarzo, line 183: tried a few variations, but allowing systems  and connections to go together seemed the most natural (also simplifies map backup/templates)

PR #389, opened 2025-05-17
fix: fixed anomalies shown as "unscanned" in ru localization but correctly filtered.

windstep:
Fixed error, when you are scanning down signature in RU localization they are correctly filtered (thanks to v1.52.4), but still showed down as "red", which means "unscanned" even they are scanned down.

TODO: gather all signature titles (which can be done easily by trying to create new filter in probes window) and create a function which will find the name in localized features


DmitryPopov:
Please update PR to use develop branch, not main, there was some signatures related changes merged today as well, probably conflicts.

windstep:
@DmitryPopov this branch could be merged, other one is under conflict resolving.

PR #416, opened 2025-05-30
fix: update zkill to use redisq instead of websocket

guarzo:
- **New Features**
  - Refactored zkillboard integration to use redisq instead of deprecated websocket functionality

- **Bug Fixes**
  - Improved error handling and logging for killmail fetching, parsing, and caching operations.
  - Enhanced reliability and observability in map and killmail data updates.

- **Refactor**
  - Replaced previous zKillboard provider modules with a more modular and maintainable architecture.
  - Simplified and unified connection creation, retrieval, and upsert logic.


lib/wanderer_app/env.ex
```diff
@@ -25,6 +25,7 @@ defmodule WandererApp.Env do
   def corp_wallet, do: get_key(:corp_wallet, "")
   def corp_eve_id, do: get_key(:corp_id, -1)
   def subscription_settings, do: get_key(:subscription_settings)
+  def endpoint_secret_key_base, do: Application.get_env(@app, WandererAppWeb.Endpoint)[:secret_key_base]
```
guarzo, line 28: @DmitryPopov unsure on this -- suggestions?

lib/wanderer_app/cached_info.ex
```diff
@@ -1,38 +1,106 @@
...
+        error  # Propagate the actual error instead of masking it
+
+      other ->
+        Logger.warning("[CachedInfo] Unexpected error fetching ship type #{type_id}: #{inspect(other)}")
+        {:error, {:unexpected_error, other}}
+    end
+  end
+
+  # Helper to fetch the group name, defaulting to "Unknown"
+  defp get_group_name(group_id) do
```
DmitryPopov, line 100: Better to move that method to esi/api_client.ex

lib/wanderer_app/map/map_zkb_data_fetcher.ex
```diff
@@ -46,6 +47,7 @@ defmodule WandererApp.Map.ZkbDataFetcher do
           end
         rescue
           e ->
+            Logger.error("[ZkbDataFetcher] Error processing map #{map_id}: #{Exception.message(e)}")
             @logger.error(Exception.message(e))
```
DmitryPopov, line 51: Get rid of this log then.

lib/wanderer_app/env.ex
guarzo: i think i was overcomplicating the queue ID -- can just do a random string on startup

lib/wanderer_app/cached_info.ex
guarzo: good call

lib/wanderer_app/esi/api_client.ex
```diff
@@ -35,7 +35,7 @@ defmodule WandererApp.Esi.ApiClient do
   @default_avoid_systems [@zarzakh_system]
 
   @cache_opts [cache: true]
-  @retry_opts [max_retries: 0, retry_log_level: :warning]
+  @retry_opts [retry: :transient, max_retries: 3, retry_log_level: :warning]
```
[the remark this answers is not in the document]
guarzo: makes sense, reversed the default

lib/wanderer_app/release.ex
```diff
@@ -19,7 +19,7 @@ defmodule WandererApp.Release do
     IO.puts("Creation of Db successful!")
   end
 
-  def migrate(repos \\ repos()) do
+  def migrate(_repos) do
```
DmitryPopov, line 22: If repos not used, better remove the param here.

lib/wanderer_app_web/live/map/event_handlers/map_kills_event_handler.ex
```diff
@@ -1,241 +1,204 @@
...
+          kills
+        {:error, _reason} ->
+          []
+      end
 
-  def handle_server_event(event, socket),
-    do: MapCoreEventHandler.handle_server_event(event, socket)
+      reply = %{kills: cached_kills}
+
+      case Task.Supervisor.start_child(WandererApp.TaskSupervisor, fn ->
```
DmitryPopov, line 87: It seems a bit complicated logic here, don't sure user page should initiate fetching kills for system, it can be several pages from different users initiate that at same time for same added system selected. Guess better to have a kind of shared GenServer to handle this logic and send loaded kills by pub/sub if system is selected on the map, and better to avoid fetching kills for same system twice (on large maps it can be an issue).

PR #422, opened 2025-06-10
feat: use external service for killmail consolidation

guarzo:


guarzo:
doing a bit more cleanup -- will finish 6/16

lib/wanderer_app/kills/client.ex
```diff
@@ -0,0 +1,457 @@
...
+    Logger.info("[Client] Retrying connection (attempt #{state.retry_count + 1}/#{@max_retries})")
+    state = %{state | retry_timer_ref: nil}
+    new_state = attempt_connection(state)
+    {:noreply, new_state}
+  end
+
+  def handle_info(:refresh_subscriptions, %{connected: true} = state) do
+    Logger.info("[Client] Refreshing subscriptions after connection")
+
+    case MapIntegration.get_all_map_systems() do
```
DmitryPopov, line 109: Is it required to get all systems from all (not deleted) maps here, or can start with running maps only?

PR #423, opened 2025-06-13
doc: updated post for notifier bot

guarzo:


DmitryPopov:
Mb better to update original post about it?

PR #432, opened 2025-06-20
feat: add versioned api and tests

guarzo:


.github/workflows/api-breaking-changes.yml
```diff
@@ -0,0 +1,134 @@
+name: API Breaking Changes Check
+
```
guarzo, line 2: sanity check to ensure we always notice breaking api  changes

.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,68 @@
+name: Code Quality
```
guarzo, line 1: this is probably overkill for now, but would be nice to get to a point where we don't have warnings

lib/wanderer_app/api/access_list_member.ex
```diff
@@ -3,13 +3,34 @@ defmodule WandererApp.Api.AccessListMember do
...
-    data_layer: AshPostgres.DataLayer
+    data_layer: AshPostgres.DataLayer,
+    extensions: [AshJsonApi.Resource]
 
   postgres do
     repo(WandererApp.Repo)
     table("access_list_members_v1")
   end
 
+  json_api do
```
guarzo, line 14: this seems like a nice way to simplify all of our crud apis

lib/wanderer_app_web/auth/auth_strategy.ex
```diff
@@ -0,0 +1,30 @@
+defmodule WandererAppWeb.Auth.AuthStrategy do
+  @moduledoc """
```
guarzo, line 2: this pattern also seems interesting, moving from the one-off plugs

lib/wanderer_app/character/tracker.ex
```diff
@@ -372,7 +371,7 @@ defmodule WandererApp.Character.Tracker do
                    access_token: access_token,
                    character_id: character_id
                  ) do
-              {:ok, ship} when is_non_struct_map(ship) ->
+              {:ok, ship} when is_map(ship) ->
```
DmitryPopov, line 374: is_non_struct_map don't allow structs here, so it was intended to use

lib/wanderer_app_web/auth/strategies/jwt_strategy.ex
```diff
@@ -0,0 +1,79 @@
+defmodule WandererAppWeb.Auth.Strategies.JwtStrategy do
```
DmitryPopov, line 1: In what scenarios we'll be needed that strategy?

PR #441, opened 2025-06-21
feat: add websocket and webhooks for events

guarzo:


lib/wanderer_app/external_events/webhook_dispatcher.ex
```diff
@@ -0,0 +1,352 @@
...
+      # Disable subscription if too many consecutive failures
+      update_attrs = if new_consecutive_failures >= @max_consecutive_failures do
+        Logger.warning("Disabling webhook subscription #{subscription.id} due to #{@max_consecutive_failures} consecutive failures")
+        Map.put(update_attrs, :active?, false)
+      else
+        update_attrs
+      end
+      
+      try do
+        MapWebhookSubscription.update!(subscription, update_attrs)
```
DmitryPopov, line 319: Do we have a way to re-activate not active sub in feature?

lib/wanderer_app_web/channels/user_socket.ex
```diff
@@ -0,0 +1,65 @@
...
+    unless WandererApp.Env.websocket_events_enabled?() do
+      require Logger
+      remote_ip = get_remote_ip(connect_info)
+      Logger.info("WebSocket connection rejected - websocket events disabled from #{remote_ip}")
+      :error
+    else
+    
+    # Extract API key from connection params
+    # Client should connect with: /socket/websocket?api_key=<key>
+    require Logger
```
DmitryPopov, line 19: Same here

lib/wanderer_app/external_events/webhook_dispatcher.ex
guarzo: via api, ui in future release?

Issue #445, opened 2025-06-27
API Modernization


guarzo:
# API Modernization: Migrate to ash_json_api with Zero Breaking Changes

## Overview
Modernize Wanderer's API by introducing a new JSON:API-compliant v1 API alongside the existing legacy API, ensuring zero breaking changes and providing a clear migration path for clients.

## Goals
- ✅ **Zero Breaking Changes**: Existing `/api/*` endpoints remain unchanged
- 🚀 **Modern Standards**: New `/api/v1/*` using ash_json_api and JSON:API spec
- 📊 **Observable**: Full OpenTelemetry instrumentation from day one
- 🔒 **Secure**: Built-in rate limiting, JWT support, and continuous security testing
- 📝 **Contract-Driven**: OpenAPI validation and consumer contract testing

### Phase 0: Quality Gates & Foundation
- [ ] Implement quality gates module to prevent regressions
- [ ] Set up lint & schema validation for JSON:API
- [ ] Configure OpenAPI spec validation in CI

### Phase 1: Comprehensive Testing
- [ ] Create integration tests for all existing API endpoints
- [ ] Document current response formats as test fixtures
- [ ] Set up end-to-end smoke tests for both legacy and v1

### Phase 2: Context Extraction
- [ ] Extract business logic from controllers into context modules
- [ ] Create clean separation between web and business layers
- [ ] Prepare codebase for dual API support

### Phase 3: ash_json_api Setup
- [ ] Add ash_json_api dependency and configuration
- [ ] Set up OpenTelemetry instrumentation
- [ ] Implement standardized error handling for JSON:API
- [ ] Configure OpenAPI auto-generation and publishing

### Phase 4: Authentication & Security
- [ ] Create unified authentication module supporting both APIs
- [ ] Add OAuth2/JWT support for modern clients
- [ ] Implement security testing (OWASP ZAP, penetration tests)
- [ ] Set up automated vulnerability scanning in CI

### Phase 5: Resource Migration
- [ ] Migrate resources in dependency order:
  - [ ] Maps (core resource)
  - [ ] MapSystems
  - [ ] MapConnections
  - [ ] MapSignatures
  - [ ] Characters
  - [ ] AccessLists
- [ ] Ensure feature parity with legacy endpoints

### Phase 6: Advanced Features 
- [ ] Implement pagination with JSON:API standards
- [ ] Add filtering and sorting capabilities
- [ ] Set up rate limiting (1000 req/min per API key)
[...]

PR #465, opened 2025-07-16
feat: implement JSON:API v1 compliant interface with versioned routing

guarzo:


guarzo:
closes #445

lib/wanderer_app/api/user_transaction.ex
```diff
@@ -3,13 +3,30 @@ defmodule WandererApp.Api.UserTransaction do
...
+  json_api do
+    type "user_transactions"
+
+    # Handle composite primary key
+    primary_key do
+      keys([:id])
+    end
+
+    routes do
+      base("/user_transactions")
```
DmitryPopov, line 23: I'm not sure we should expose the transactions API

PR #477, opened 2025-07-27
feature: provide default settings interface

guarzo:


DanSylvest:
@guarzo 
image
For user default settings. You need use style of this button

DanSylvest:
Next: Behavior of settings should be next.
We must provide
1. Reset of common Map settings not influent on your personal settings. (This button SHOULD NOT remove personal settings of admin if on server side wasn't found config).
1.1. For this action you need to show popup (link in comments remove, which ask really we want to do it)
2. We SHOULD show somewhere (in Admin tab), what we have some Map settings (if it already applied to Map) and show and when add it.
3. We MUST give ability to any user decline using Map default settings. It means that somewhere should be button which allow him do it.

PR #509, opened 2025-09-01
fix: ensure pub/sub occurs after acl api change

guarzo:


lib/wanderer_app_web/controllers/access_list_member_api_controller.ex
```diff
@@ -385,13 +395,27 @@ defmodule WandererAppWeb.AccessListMemberAPIController do
                    :acl_member_removed
                  ) do
               :ok ->
+                # Also broadcast internal PubSub message for map servers
+                Phoenix.PubSub.broadcast(
```
DmitryPopov, line 399: use broadcast_acl_updated(acl_id) here

PR #511, opened 2025-09-02
feat(Map): Add "locked and chains" scope

erais-bereund:
# What problem does this attempt to solve?

I'm living in nullsec. We use Wanderer to track wormhole chains originating from our space. This map has our home space locked.

We have a policy for players to manually disable tracking when traveling through k-space (outside our home space).

We do this to keep the map somewhat 'clean', but occasionally a player forgets to turn tracking off. No big deal, someone cleans it up, but it still takes time away from other things.

# How does this solve the problem?

By introducing a new map scope `locked and chains` which tracks connections between

1. wormhole to wormhole space (chains)
2. wormhole to known space (chain exits)
3. locked systems to anywhere (chain entries and direct k-space connections)

we can abolish our policy. Players can just keep tracking enabled for such a map, and will not pollute the map with k-space systems we aren't interested in.

# Known issue: 'side' chains

This method isn't perfect as it still creates 'side' chains that are not connected to any locked systems.

I did not check if there was functionality to find a route through the map, mostly because I didn't want to introduce a graph-search each time the check is run. If that is possible and not too costly, I would change the code to use 'the map contains a connection to a locked system' for either the `from` or the `to` of a new connection as the sole criterion.

The initially described behavior is however good enough for visits to trade hubs.

# Alternatives

I'm pretty sure this could also be implemented as a 'tracking modifier' somewhere in the settings (i.e., independent from the map scope).

# Other stuff

This is my first time working with Elixir, so please mention any nonsense I produced :laughing: Apart from that I'd just wanted to mention that you have a codebase that is very easy to read and follow! It was a pleasure to hack on it!


DmitryPopov:
'Nullsec' scoped maps will auto-add N-sec systems only + 1 border LS/WH/HS system, as same as WH scope works actually.

PR #517, opened 2025-09-16
Show system temp name when linking signatures

jackmurray:
This PR enables the display of a system's temporary name on the signature linking dropdown menu.

I'm by no means a good react developer, and I've been looking at the Wanderer codebase for about 30 minutes before making this so I'm sure it needs more work, I just don't know what that is.

If a maintainer could look at this and let me know where to go from here that'd be great.

### DIFF
--- a/assets/js/hooks/Mapper/components/ui-kit/SystemView/SystemView.tsx
+++ b/assets/js/hooks/Mapper/components/ui-kit/SystemView/SystemView.tsx
@@ -42,5 +42,5 @@ export const SystemView = ({ systemId, systemInfo: customSystemInfo, showCustomN
     return <SystemViewStandalone {...rest} {...systemInfo} />;
   }
 
-  return <SystemViewStandalone customName={mapSystemInfo.name ?? undefined} {...rest} {...systemInfo} />;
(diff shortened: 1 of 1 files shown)


DanSylvest (changes requested):
flag showCustomName using either in another places what leads to show temporary_name everywhere - not only in this widget. 
First of all you should explain what you want to get. Because this code i can't to accept

jackmurray:
Basically what I want to do is show the temporary name of a system, if it has one, instead of the regular system name.

The reason being that if you've set a temp name, you should see/interact with the system using that name instead of the real name/J-number.

DanSylvest:
> Basically what I want to do is show the temporary name of a system, if it has one, instead of the regular system name.
> 
> The reason being that if you've set a temp name, you should see/interact with the system using that name instead of the real name/J-number.

Problem here - that we already have custom name and temporary name will override it. Conceptually it means that we should see either temp and custom. Because both names matter.

DmitryPopov:
I guess system names showed on link signature to system modal should be affected only. So if priority for names are: temp > custom > j-name, what if we show systems names for link signature modals as: 'temp (custom/j-name)'?

jackmurray:
Sure, that works.

Issue #492, opened 2025-08-07
Allow setting the same owner on multiple structure at a time


sinulf:
Right now, the name of the owner needs to be typed manually each time, then clicked in the matching results. Please allow setting the owner once for multiple selected structures to speed things up (this is possible in Pathfinder).

PR #530, opened 2025-10-19
feat(map): Update Owners on Multiple Structures

s-no1ukno:
This is a pull request in response to Issue #492.

I ended up using a modified instance of the same modal that was being used to update structures to modify owners on multiple structures simultaneously.  The new modal is triggered by clicking a new icon on the top of the System Structures widget - this icon only appears if there's more than one structure in system.  

This probably needs a little polish in the form of CSS (I'm terrible at styling).  If anyone wanted to handle that, it would be much appreciated.

### DIFF
--- a/assets/js/hooks/Mapper/components/mapInterface/widgets/SystemStructures/SystemStructures.tsx
+++ b/assets/js/hooks/Mapper/components/mapInterface/widgets/SystemStructures/SystemStructures.tsx
@@ -15,6 +15,7 @@ import { SystemStructuresContent } from './SystemStructuresContent/SystemStructu
 import { useSystemStructures } from './hooks/useSystemStructures';
 import { processSnippetText, StructureItem } from './helpers';
 import { SystemStructuresOwnersDialog } from './SystemStructuresOwnersDialog/SystemStructuresOwnersDialog';
+import clsx from 'clsx';
 
 export const SystemStructures: React.FC = () => {
   const {
@@ -47,9 +48,16 @@ export const SystemStructures: React.FC = () => {
   );
 
   const handleSave = (updatedStructures: StructureItem[]) => {
-    console.log("calling handleUpdateStructures now")
     handleUpdateStructures(updatedStructures)
   }
+
+  const handleOpenDialog = useCallback(() => {
+    setShowEditDialog(true)
+  }, [])
+
+  const handleCloseDialog = useCallback(() => {
+    setShowEditDialog(false)
+    }, [])
 
   const handlePasteTimer = useCallback(async () => {
     try {
@@ -76,8 +84,8 @@ export const SystemStructures: React.FC = () => {
         <LayoutEventBlocker className="flex gap-2.5">
           {structures.length > 1 && (
             <WdImgButton
-              className={`${PrimeIcons.USER_EDIT} text-sky-400 hover:text-sky-200 transition duration-300`}
-              onClick={() => setShowEditDialog(true)}
+              className={clsx(PrimeIcons.USER_EDIT, 'text-sky-400 hover:text-sky-200 transition duration-300')}
+              onClick={handleOpenDialog}
               tooltip={{
                 position: TooltipPosition.left,
                 // @ts-ignore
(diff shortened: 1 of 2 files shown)


assets/js/hooks/Mapper/components/mapInterface/widgets/SystemStructures/SystemStructures.tsx
```diff
@@ -44,6 +46,11 @@ export const SystemStructures: React.FC = () => {
     [processClipboard],
   );
 
+  const handleSave = (updatedStructures: StructureItem[]) => {
+    console.log("calling handleUpdateStructures now")
+    handleUpdateStructures(updatedStructures)
+  }
+
```
s-no1ukno: Fixed this `console.log()` statement, missed it on my first pass through.  It'll be removed on the review commit.

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemStructures/SystemStructuresOwnersDialog/SystemStructuresOwnersDialog.tsx
```diff
@@ -0,0 +1,152 @@
...
+        setOwnerSuggestions(results);
+        setPrevQuery(newQuery);
+        setPrevResults(results);
+      } catch (err) {
+        console.error('Failed to fetch owners:', err);
+        setOwnerSuggestions([]);
+      }
+    },
+    [prevQuery, prevResults, outCommand],
+  );
```
s-no1ukno: This file started as an exact copy of the other structure edit dialog box - copying and then cutting out bloat for the new dialog box seemed like a better option than modifying the original dialog to have a bunch of logic checks and modes.  The problem is there is some repeated code now.  This could be cleaned up, for example moving the `searchOwners()` function above into a helper file.  Just wasn't sure of details such as where to put it and whether that was most ideal.

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemStructures/SystemStructures.tsx
```diff
@@ -67,6 +74,17 @@ export const SystemStructures: React.FC = () => {
         </div>
 
         <LayoutEventBlocker className="flex gap-2.5">
+          {structures.length > 1 && (
+            <WdImgButton
+              className={`${PrimeIcons.USER_EDIT} text-sky-400 hover:text-sky-200 transition duration-300`}
```
DanSylvest: Please use 'clsx' method for concat classes 
```
clsx(PrimeIcons.USER_EDIT, 'text-sky-400 hover:text-sky-200 transition duration-300')
```

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemStructures/SystemStructuresOwnersDialog/SystemStructuresOwnersDialog.tsx
```diff
@@ -0,0 +1,158 @@
...
+      return { ...item, ownerName: selected.label, ownerId: selected.value }
+    }))
+  };
+
+  const handleSaveClick = async () => {
+    if (!editData) return;
+
+    // fetch corporation ticker if we have an ownerId
+    for (const structure of editData) {
+      if (structure.ownerId) {
```
DanSylvest: first of all never do any requests inside of loop.
next - always use exit-first patter. Condition - `if (structure.ownerId) {` should be reversed as `if (!structure.ownerId) {` and leads to continue.

Second. You need to use Promise.all for all requests which can be exectued in one time. 
for example

```
const allOwnerIds = editData.filter(x => x.ownerId != null).map(x => x.ownerId)
const tickers = await Promise.all(allOwnerIds.map(id => getOwnerId(id)))
// ...then
```

 you get index by each request and take ownerId from allOwnerIds then find in editData by ownerId data and update it. But as I see - you mutating editData what is incorrect way. You need to update it. What mean you should create new array with new data with updated values. After it call save

PR #574, opened 2026-01-14
Map Auto Layout

fbnyes:
Short summary
• auto layout
• auto numbering
• cycle detection inside a tree
• marking when two separate trees get connected (red marking still a bit scuffed)
• settings added for all relevant new options

Other stuff
• unsplashed sigs
• text input when updating a sig

I put up what I have right now, normally I would look into separating stuff - but I would rather take this PR to gauge interest and if the way it is done is something that could work longtime.

If we agree on a solution I would be happy to implement and post a neat PR. 

o7

### DIFF
--- a/assets/js/hooks/Mapper/components/map/components/ContextMenuRoot/useContextMenuRootHandlers.ts
+++ b/assets/js/hooks/Mapper/components/map/components/ContextMenuRoot/useContextMenuRootHandlers.ts
@@ -46,6 +46,22 @@ export const useContextMenuRootHandlers = ({ onAddSystem, onCommand }: UseContex
     ref.current.onAddSystem?.({ coordinates: position });
   }, [position]);
 
+  const onAutoLayout = useCallback(async () => {
+    const { onCommand } = ref.current;
+    if (!onCommand) {
+      return;
+    }
+
+    console.log('Auto layouting systems');
+
+    await onCommand({
+      type: OutCommand.layoutSystems,
+      data: {
+        system_ids: [], // Layout all systems
+      },
+    });
+  }, []);
+
   const onPasteSystemsAnsConnections = useCallback(async () => {
     const { pasteSystemsAndConnections, onCommand, position } = ref.current;
     if (!position || !onCommand || !pasteSystemsAndConnections) {
@@ -72,5 +88,6 @@ export const useContextMenuRootHandlers = ({ onAddSystem, onCommand }: UseContex
     contextMenuRef,
     onAddSystem: onAddSystemCallback,
     onPasteSystemsAnsConnections,
+    onAutoLayout,
   };
 };
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -1,0 +1,12939 @@
+{
+  "name": "assets",
+  "version": "1.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "assets",
+      "version": "1.0.0",
+      "license": "ISC",
+      "dependencies": {
+        "@codemirror/lang-markdown": "^6.3.2",
+        "@codemirror/theme-one-dark": "^6.1.2",
+        "@formkit/auto-animate": "0.7.0",
+        "@shopify/draggable": "^1.1.3",
+        "@uiw/react-codemirror": "^4.23.10",
+        "clsx": "^2.1.1",
+        "daisyui": "^4.11.1",
+        "fast-deep-equal": "^3.1.3",
+        "live_select": "file:../deps/live_select",
+        "lodash.debounce": "^4.0.8",
+        "lodash.isequal": "^4.5.0",
+        "phoenix": "file:../deps/phoenix",
+        "phoenix_html": "file:../deps/phoenix_html",
+        "phoenix_live_view": "file:../deps/phoenix_live_view",
+        "primeflex": "^3.3.1",
+        "primeicons": "^7.0.0",
+        "primereact": "10.6.5",
+        "prism-themes": "^1.9.0",
+        "react": "^18.3.1",
+        "react-dom": "^18.3.1",
+        "react-error-boundary": "^4.0.13",
+        "react-event-hook": "^3.1.2",
+        "react-flow-renderer": "^10.3.17",
+        "react-hook-form": "^7.53.1",
+        "react-markdown": "^10.0.1",
+        "react-transition-group": "^4.4.5",
+        "react-usestateref": "^1.0.9",
+        "reactflow": "^11.11.4",
+        "remark-breaks": "^4.0.0",
+        "remark-gfm": "^4.0.1",
+        "tailwindcss": "^3.3.6",
+        "topbar": "^3.0.0",
+        "use-local-storage-state": "^19.3.1"
+      },
+      "devDependencies": {
+        "@tailwindcss/aspect-ratio": "^0.4.2",
+        "@tailwindcss/forms": "^0.5.7",
+        "@tailwindcss/typography": "^0.5.13",
+        "@types/jest": "^29.5.12",
+        "@types/lodash.debounce": "^4.0.9",
+        "@types/lodash.isequal": "^4.5.8",
+        "@types/react": "^18.3.12",
+        "@types/react-dom": "^18.3.1",
+        "@typescript-eslint/eslint-plugin": "^6.21.0",
+        "@typescript-eslint/parser": "^6.21.0",
+        "@vitejs/plugin-react": "^4.3.3",
+        "@vitejs/plugin-react-refresh": "^1.3.6",
+        "autoprefixer": "^10.4.19",
+        "babel-jest": "^29.7.0",
+        "child_process": "^1.0.2",
+        "eslint": "^8.57.0",
+        "eslint-config-prettier": "^9.1.0",
+        "eslint-plugin-prettier": "^5.1.3",
+        "eslint-plugin-react": "^7.34.1",
+        "eslint-plugin-react-hooks": "^4.6.0",
+        "eslint-plugin-react-refresh": "^0.4.6",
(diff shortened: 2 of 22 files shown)


guarzo (commented):
Read on the approach, since I'd like this feature to exist. Auto layout and auto numbering are things people currently do by hand and get wrong, and the algorithm itself reads clearly — the component finding and oldest-connection-first tree building are easy to follow. 283 lines of unit tests on something you called a rough cut is more than most merged PRs here manage, and it's what makes the layout logic reviewable at all.

The thing I'd settle before writing the neat version isn't the layout maths, it's **when it runs**. `auto_layout` hangs off `maybe_add_connection`, which is on the character-movement path, so every jump that opens a new connection relayouts the whole map inside the map GenServer — a read-then-write per system plus a write and broadcast per connection. That's heavy on a hot path, and it moves systems under people mid-drag. Details inline, along with a few smaller things.

Two practical blockers. It targets **`develop`**, which has been stale since Feb — `main` is the default now and 184 commits ahead, so this can't merge as-is. And `assets/package-lock.json` is committed: 12,939 of the 14,729 added lines. This repo is yarn, so it looks like a stray `npm install`. Dropping it takes the diff to ~600 lines of actual code.

Smaller: `mix format` fails on `map_position_calculator.ex` and `map_server_connections_impl.ex` (both clean on the base). And since it adds a context menu item and restructures the settings modal, screenshots would help — particularly for the cross-tree marking you mentioned is still scuffed.

lib/wanderer_app/map/server/map_server_connections_impl.ex
```diff
@@ -646,6 +666,13 @@ defmodule WandererApp.Map.Server.ConnectionsImpl do
           solar_system_target_id: location.solar_system_id
         })
 
+        if retrigger_layout do
```
guarzo, line 669: This is the one I'd most want settled before you write the tidy version.

`maybe_add_connection/7` is called from `map_server_characters_impl.ex:968` — the character tracking path. So with `auto_layout` on, every time anyone on the map jumps through a hole that isn't already mapped, this fires `SystemsImpl.layout_systems(map_id, nil)`, which relayouts **the entire map**, not just the affected subtree.

Costwise, per trigger: `layout_systems` calls `update_system_position` per system, and each of those does a `get_by_map_and_solar_system_id` read followed by a write, plus an rtree update and a cache put. Then it walks every connection and, for any whose `custom_info` changed, does another write plus an `Impl.broadcast!`. On the first run every connection changes because the key is new, so that's the full set. For a 150-system chain during a fleet op with several people jumping around, that's a lot of synchronous work in the map GenServer, which is also serving everything else on that map.

There's a UX half to this too, and it might matter more: systems move under people while they're using the map. Someone dragging a system into place gets it yanked elsewhere because a fleetmate jumped somewhere unrelated.

Options that seem worth weighing, roughly in order of how much I'd like them:

- Only relayout the affected component rather than the whole map — you already compute connected components, so the information is right there.
- Make it explicit-only (the context menu item you've already built) and drop the automatic trigger entirely. Least code, and arguably what most people want anyway.
- Keep it automatic but debounce it in the map server so a burst of jumps collapses into one relayout.

The `retrigger_layout` flag threaded through `add_connection`/`maybe_add_connection` is a reasonable way to stop `paste_connections` triggering N relayouts, so the instinct is right — I just think the problem it's patching goes away if the trigger moves.

lib/wanderer_app/map/map_position_calculator.ex
```diff
@@ -119,4 +119,325 @@ defmodule WandererApp.Map.PositionCalculator do
...
+          acc
+          |> Map.update(s, [t], &[t | &1])
+          |> Map.update(t, [s], &[s | &1])
+        else
+          acc
+        end
+      end)
+
+    # 4. Layout each root sequentially
+    layout_type = (opts |> Keyword.get(:layout, "left_to_right")) |> String.to_atom()
```
guarzo, line 172: `String.to_atom` on this one is worth changing whichever way the rest goes.

`opts[:layout]` comes from `map_options/1` (`map_server_impl.ex:474`), which reads `options["layout"]` straight out of the map's stored options. Those come from the settings form via `Map.take` in `maps_live.ex:568` with no validation of the value — so this is a user-supplied string reaching `String.to_atom`, and atoms are never garbage collected. Anyone who can edit map settings can grow the atom table permanently.

The neat part is the fix is already sitting in both files. `map_options/1` uses `String.to_existing_atom` for every boolean right above this, and this very file pattern-matches the layout as a plain string 70-odd lines up:

```elixir
defp get_start_index(n, "left_to_right"), do: div(n, 2)
defp get_start_index(n, "top_to_bottom"), do: div(n, 2) + n - 1
```

Matching on the strings the same way would drop the conversion entirely and keep the file internally consistent.

Unrelated but nearby: `Logger.info` on every layout call, plus one line per special connection, is going to be noisy at info level if this ever runs automatically. `Logger.debug` seems more like it.

lib/wanderer_app/map/server/map_server_systems_impl.ex
```diff
@@ -283,6 +283,72 @@ defmodule WandererApp.Map.Server.SystemsImpl do
...
+      )
+
+    show_temp_system_name = map_opts |> Keyword.get(:show_temp_system_name, false)
+
+    updated_systems
+    |> Enum.each(fn updated_system ->
+      hierarchical_name = Map.get(updated_system, :hierarchical_name)
+
+      if show_temp_system_name and not is_nil(hierarchical_name) and hierarchical_name != "0" do
+        update_system_temporary_name(map_id, %{
```
guarzo, line 315: Auto-numbering writes the hierarchical name into `temporary_name`, which is the same field people type into by hand — so a relayout silently overwrites whatever someone had named a system, with no undo. Temp names are often doing real work on a chain ("HOME", "camp", "bring dps"), so losing them to a background relayout would sting.

It's also gated on `show_temp_system_name`, which today means "allow temporary system names" — a permission-ish toggle about whether the feature exists at all. Reusing it as "and also let auto-layout rename things" means anyone who wants manual temp names gets automatic renaming whether they asked for it or not. Given you've already added `auto_layout` as its own option, a separate `auto_numbering` toggle seems like the natural shape.

If the numbering is worth keeping (I think it is — it's one of the nicer bits here), storing it in its own field rather than borrowing `temporary_name` would let both exist, and would let the frontend show the hierarchy without destroying anything.

lib/wanderer_app_web/live/map/map_event_handler.ex
```diff
@@ -339,19 +340,34 @@ defmodule WandererAppWeb.MapEventHandler do
...
-        mass_status: mass_status,
-        time_status: time_status,
-        type: type,
-        ship_size_type: ship_size_type,
-        locked: locked,
-        source: "#{solar_system_source}",
-        target: "#{solar_system_target}"
-      }
+      ) do
+    is_cross_list =
```
guarzo, line 347: Two things about carrying `is_cross_list` inside `custom_info` JSON.

This decode now runs for **every connection, every time connections are serialised to the frontend** — so it's on the map render and broadcast path, not just on layout. It's cheap per call but it's the kind of thing that's annoying to unpick later.

More the design question: `is_cross_list` is derived layout state, and `custom_info` is a free-form blob shared with whatever else wants to write there. The merge is careful (you decode, `Map.put`, re-encode, and only write when it actually changed, which does preserve other keys — that part's right), but it means a boolean is stored as a substring of a JSON text column, can't be queried, and quietly disagrees with reality the moment someone adds a connection without a relayout following.

A real `boolean` attribute on `MapConnection` would be cheaper on the read path, visible in the resource, and would make the staleness question explicit. Adds a migration and a snapshot regen, but it's the sort of thing that's much easier to do now than after it ships.

(Small thing on the same block: `conn.id in cross_list_conn_ids` at `map_server_systems_impl.ex:330` is a list scan inside a loop over all connections, so it's quadratic. A `MapSet` would make it not. Irrelevant at 50 connections, less so at 400.)

assets/js/hooks/Mapper/components/map/components/ContextMenuRoot/useContextMenuRootHandlers.ts
```diff
@@ -46,6 +46,22 @@ export const useContextMenuRootHandlers = ({ onAddSystem, onCommand }: UseContex
     ref.current.onAddSystem?.({ coordinates: position });
   }, [position]);
 
+  const onAutoLayout = useCallback(async () => {
+    const { onCommand } = ref.current;
+    if (!onCommand) {
+      return;
+    }
+
+    console.log('Auto layouting systems');
```
guarzo, line 55: `console.log('Auto layouting systems')` wants to come out — worth a grep of the branch for others before the tidy version.

While I'm in this file: the "Auto Layout" menu item isn't permission-gated on the frontend, but the LiveView handler is — `map_systems_event_handler.ex` matches on `user_permissions: %{update_system: true}`, so someone without it clicks the item and *nothing happens at all*, no error, no toast. Not a security problem (the server does hold the line), just a dead button.

The Paste item immediately below yours already solves exactly this: it's `disabled: !allowPaste` with a `MenuItemWithInfo` tooltip explaining why. Copying that pattern for Auto Layout would make it consistent and would tell people what's wrong.

assets/package-lock.json
```diff
@@ -0,0 +1,12939 @@
+{
```
guarzo, line 1: I think this got committed by accident — the repo is yarn (there's a `yarn.lock`, and `make yarn` / the docs all use yarn), and an `npm install` will happily generate this alongside it.

It's 12,939 of the 14,729 added lines in this PR. Deleting it, plus reverting the `yarn.lock` regeneration if nothing actually needed a dependency change, would take the diff down to roughly 600 lines of real code.

Worth checking whether `assets/.gitignore` covers `package-lock.json` too, so it can't come back.

Issue #600, opened 2026-03-21
C729 changes for patch 2026-03-18.1


starkythefox:
_(Since the patch for the lifetimes and the new WHs have been introduced in a merge, I'm writing this issue for the C729 ones)_

Relevant info only for the C729 changes

> - C729 wormholes must now be scanned down within Pochven and will connect out to k-space.
> - The Reliable Lifetime of all extra-Pochven wormholes has been adjusted:
>   - The timer has increased from 6 to 12 hours for C729 wormholes.

In other words:

- C729 and its K162 have been reversed: C729 in Pochven, K162 where they would have been C729 instead (ej: Shihuken for Senda)
- C729 are now static, with source system in Pochven and destination systems to HS, LS, NS and I assume Pochven systems still (ex: Ahtila and Kino had pre-change C729s spawn in Pochven systems)
- The lifetime change either makes no sense or Wanderer has been incorrect until now, but Hoboleaks shows no lifetime changes for C729. In any case, Wanderer uses 12h already.

So my changes would have been made here in `priv/repo/data/wormholes.json`:
```diff
@@ -177,14 +177,14 @@
   {
     "mass_regen": 0,
-    "dest": "pochven",
-    "src": ["hs", "ls", "ns", "pochven"],
-    "static": false,
+    "dest": ["hs", "ls", "ns", "pochven"],
+    "src": "pochven",
+    "static": true,
     "max_mass_per_jump": 1000000000,
     "lifetime": "12",
     "total_mass": 1000000000,
     "name": "C729",
-    "respawn": ["wandering", "reverse"]
+    "respawn": ["static"]
   },
```

But since I don't see any wormhole that has `dest` as array, and I can't get to setup a local env for wanderer to test if anything broke (wormhole selection, system selection during wormhole/connection creation/edition, wormhole list info on the right panel, ...), I cannot assure above will not create errors.

PR #603, opened 2026-04-12
fix: C729 reversal - Allow multiple destinations for Wormholes

starkythefox:
The main objective of this PR was the reversal of C729 wormholes by making them static on Pochven and outgoing to K-space and Pochven systems as of patch 2026-03-18.1:
imagen

But due to the fact that, outside of the special case of K162, no wormhole has ever lead to multiple system classes, it also meant reworking the application to allow just that. It closes my issue here: https://github.com/wanderer-industries/wanderer/issues/600; as well as an issue and FR in Discord for the same

The main changes of this PR:
- Make all wormholes accept an array of class destinations.
  - For All but C729 and K162, I just made the previous destination be an array of said single element.
  - For K162, its value of `null` stays to help keeping its old behavior.
- Reverse C729 `src` and `dest` values, mark as `static` and replace `"reverse"`and `"wandering"` for just `"static"`.
- Make C729 static to all 27 Pochven systems
- Add the "K162 Type" field to C729 types during Signature edition. Renamed to "Destination Class" and as well as name files and variables to make it more generic for any future wormhole with multiple destinations:
  - For K162, no changes.
  - For C729, only K-space security classes and Pochven are listed
- During signature edition, if multiple systems are connected, "Leads to" will filter for only allowed outgoing connections (K-space or Pochven)
- As with K162 wormholes:
  - C729s will show the destination class appended to its WH class name (ex: C729 High-Sec) on the system signatures panel when no target system is linked yet to it.
  - C729s unsplashed signature will appear with the selected target class color if set; or default color if none.
- Signature linking, by connecting from anchor to anchor or when the WH jump is tracked, will filter correctly depending if "Destination Class" is set on C729 (only show on matching-class systems) or not (any K-space + Pochven)
  - Similar to K162. If C729 has High-Sec defined and linked system is Low-Sec it won't appear on the list. If it was High-Sec it would be.
- Wormhole References dialog has the updated info for C729, as in "Leads to Pochven", "static".

If there's anything you feel it's missing, or any changes you want, feel free to tell.

Screenshots:
- C729 shows as static:
imagen

- Renamed "K162 Type:" field shows when type is C729 as well:
imagen

- Various wromhole tests with different types, target class, lead to system, EOL/mass status:
imagen

- All C729 cases in signature panel and visualization on map:
imagen

- Linking between anchors from Pochven to Hi-Sec system, showcasing the filtering:
imagen

- Same but to a C4 WH, all C729 are filtered out:
imagen

- Even with C2 connected, only linked Null-Sec and Pochven appear in "Leads To:" for C729:
imagen

- Updated Wormhole Reference info for C729:
imagen

### DIFF
--- a/assets/js/hooks/Mapper/components/map/constants.ts
+++ b/assets/js/hooks/Mapper/components/map/constants.ts
@@ -314,6 +314,13 @@ export const WORMHOLES_ADDITIONAL_INFO_RAW: WormholesAdditionalInfoType[] = [
     title: 'Reverse',
     shortTitle: 'K162',
   },
+  {
+    id: 'c729',
+    shortName: 'C729',
+    wormholeClassID: 10102,
+    title: 'Static',
+    shortTitle: 'C729',
+  },
 ];
 
 export const WORMHOLES_ADDITIONAL_INFO: Record<string, WormholesAdditionalInfoType> =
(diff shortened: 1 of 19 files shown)


DanSylvest:
Hi! Good job. Thanks. Looks pretty good

PR #609, opened 2026-04-27
Fix: Update containers to successfully run

Drewsif:
When using Docker Compose, the container was failing to connect to the DB server. The two issues were the Elixir dev script was not pulling the hostname from the environment variable set and it also looked like the setup script used netcat, which was missing from the docker container so I added that in as well. Small changes but made docker compose work for me.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -4,7 +4,7 @@ import Config
 config :wanderer_app, WandererApp.Repo,
   username: "postgres",
   password: "postgres",
-  hostname: "localhost",
+  hostname: System.get_env("DB_HOST", "localhost"),
   database: "wanderer_dev",
   stacktrace: true,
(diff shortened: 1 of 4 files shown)


config/dev.exs
```diff
@@ -4,7 +4,7 @@ import Config
 config :wanderer_app, WandererApp.Repo,
   username: "postgres",
   password: "postgres",
-  hostname: "localhost",
+  hostname: System.get_env("DB_HOST", "localhost"),
```
DmitryPopov: dev config isn't a best place to read env variables, to reduce libs compilation on each server start I'd recommend to find a way to use runtime config to read it if possible

Drewsif: Moved it to `config/runtime.exs` to read the `DB_HOST` environment variable at runtime for the `dev` and `test` environments.

PR #623, opened 2026-07-12
Feature kill notification

Firepoison:


Firepoison:
This feature adds a notification toast that brings up kills that have happened in the active chain in map, alongside a sound as well.

Adds options for the user to opt in or out of the notifications as well.

guarzo:
Nice, I've wanted kill notifications on the map for a while. Had a read through and left a few notes below, mostly small stuff. 

Also, could you throw a screenshot or two in the description? The toast and the new checkboxes in settings would be enough.

assets/js/hooks/Mapper/mapRootProvider/constants.ts
```diff
@@ -23,6 +23,8 @@ export const STORED_INTERFACE_DEFAULT_VALUES: InterfaceStoredSettings = {
   pingsPlacement: PingsPlacement.rightTop,
   minimapPlacement: MiniMapPlacement.rightBottom,
   hideBookmarkWarning: false,
+  killActivityNotifications: false,
+  killActivitySounds: false,
```
guarzo, line 27: I think this needs a settings migration to go with it (same for the matching keys in `mapRootProvider/types.ts`). Adding keys to `STORED_INTERFACE_DEFAULT_VALUES` works fine for a fresh browser, but anyone with an existing `storedSettings` object in localStorage won't have these keys, so `interfaceSettings.killActivityNotifications` comes back `undefined` for them. Falsy, so it won't crash, but it also means the checkboxes will read wrong until they toggle them.

Fix is bumping `STORED_SETTINGS_VERSION` in `mapRootProvider/version.ts` and adding a `to_5` under `migrations/list/` (registered in `list/index.ts`). Two new variables = minor bump per DanSylvest's scheme.

assets/js/hooks/Mapper/components/mapRootContent/components/MapKillsNotifications/MapKillsNotifications.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+      // Find new kills we haven't seen before
+      const newKills = kills.filter(k => {
+        if (!k.killmail_id || seenKillmailIdsRef.current.has(k.killmail_id)) {
+          return false;
+        }
+        if (k.kill_time) {
+          const killTime = new Date(k.kill_time).getTime();
+          const now = Date.now();
+          if (now - killTime > 15 * 60 * 1000) {
```
guarzo, line 64: Could this get a named constant at the top of the file? Something like `const KILL_FRESHNESS_MS = 15 * 60 * 1000;`. Same for the `life: 5000` further down.

Small related thing: kills without a `kill_time` skip this check entirely and always notify. Probably fine in practice, just flagging in case it isn't intentional.

assets/js/hooks/Mapper/components/mapRootContent/components/MapKillsNotifications/MapKillsNotifications.tsx
```diff
@@ -0,0 +1,122 @@
...
+        const tempName = systemOnMap.temporary_name || systemOnMap.name;
+        const systemName = tempName && tempName !== staticName ? `${tempName} (${staticName})` : staticName;
+        
+        for (const kill of newKills) {
+          const victimName = kill.victim_char_name || 'Unknown Pilot';
+          const shipName = kill.victim_ship_name || 'Unknown Ship';
+          
+          toastRef.current?.show({
+            severity: 'warn',
+            summary: `Kill in ${systemName}`,
```
guarzo, line 94: This fires one toast per kill. If someone drops a fleet in a system on your chain you'd get a stack of them all at once. Would it make sense to collapse per system when there's more than one, e.g. "3 kills in Jita"?

assets/js/hooks/Mapper/components/mapRootContent/components/MapKillsNotifications/MapKillsNotifications.tsx
```diff
@@ -0,0 +1,122 @@
...
+      }
+
+      if (interfaceSettings.killActivitySounds) {
+        shouldPlaySound = true;
+      }
+    }
+
+    if (shouldPlaySound && audioRef.current) {
+      audioRef.current.currentTime = 0;
+      audioRef.current.play().catch(e => console.warn('Could not play kill sound:', e));
```
guarzo, line 108: `console.warn` here will get flagged in review — @DmitryPopov  hates log noise. Autoplay rejection is expected on first load anyway (browsers block it until the user has interacted with the page), so I'd just swallow it with `.catch(() => {})`.

assets/js/hooks/Mapper/components/mapRootContent/components/MapKillsNotifications/MapKillsNotifications.tsx
```diff
@@ -0,0 +1,122 @@
...
+      return true;
+    }
+
+    let shouldPlaySound = false;
+
+    for (const [systemId, kills] of Object.entries(event.data)) {
+      // Check if this system is on our map
+      const systemOnMap = systems.find(
+        s => s.id === systemId || s.system_static_info?.solar_system_id?.toString() === systemId
+      );
```
guarzo, line 53: Minor perf thing: this does a `systems.find` for every system in the event payload, and the comparison stringifies `solar_system_id` each time. Building a `Map` keyed by both `id` and the stringified solar system id once (in a `useMemo` on `systems`) would avoid the repeated scans.

Also `seenKillmailIdsRef` grows for the lifetime of the session and never gets pruned. Not urgent, but on a long session with an active chain it'll add up.

lib/wanderer_app_web/components/layouts/root.html.heex
```diff
@@ -33,6 +33,7 @@
       type="font/woff2"
       crossorigin="anonymous"
     />
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@20..48,100..700,0..1,-50..200" />
```
guarzo, line 36: Is the whole Material Symbols variable font worth it for one skull glyph? It's a fair chunk of download, and it's a render-blocking stylesheet from a third party. We already have PrimeIcons loaded (the line you're replacing used `PrimeIcons.BOLT`) and the `hero-*` set is used elsewhere in the app — is there something in either that would do? Otherwise an inline SVG would avoid the network dependency entirely.

assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.tsx
```diff
@@ -50,8 +50,8 @@ export const SolarSystemNodeDefault = memo((props: NodeProps<MapSolarSystemType>
               killsActivityType={localKillsActivityType}
               className={clsx(classes.Bookmark, MARKER_BOOKMARK_BG_STYLES[localKillsActivityType])}
             >
-              <div className={clsx(classes.BookmarkWithIcon)}>
-                <span className={clsx(PrimeIcons.BOLT, classes.icon)} />
+              <div className={clsx(classes.BookmarkWithIcon, 'gap-[2px]')}>
+                <span className={clsx('material-symbols-outlined', classes.icon)} style={{ fontSize: '9px', lineHeight: 1 }}>skull</span>
```
guarzo, line 54: Two small things here (same change is in `SolarSystemNodeTheme.tsx`): the inline `style={{ fontSize: '9px', lineHeight: 1 }}` would be better in `classes.icon` alongside the rest of the node styling, and this line is over Prettier's 120 char limit so `yarn format` will want to reflow it.

Worth running Prettier over the branch generally — there's some trailing whitespace in `MapKillsNotifications.tsx` too (lines 18, 21, 68, 71, 87, 91).

DanSylvest:
Hi! @Firepoison - i've added some rules for PR. Please read it and update your PR. https://discord.gg/UF6hfHKe9

Regarding your changes - please attach images how it looks like

PR #625, opened 2026-07-21
feat: autofocus on signature descryption field when doubleclicked/editing

SzesciennaGeometria:
small quality of life change: when doubleclicking/editing pasted signature "type descryption" is automaticaly focused and ready to receive keyboard input 
(1 less click required - based on system settings windows when doubleclicking system on the map)

edit: oj i gues i dont know how to use git properly yet so it goes into same pull-request i gues:
also increased limit for custom label to 11 characters hear me out - i tried to break the interface and you literaly have to have ALL the label thingies up for it to start poping out to the right side of the system. i think having few more characters to use will outweight that one edge-case.

### DIFF
--- a/assets/js/hooks/Mapper/components/mapInterface/components/SystemCustomLabelDialog/SystemCustomLabelDialog.tsx
+++ b/assets/js/hooks/Mapper/components/mapInterface/components/SystemCustomLabelDialog/SystemCustomLabelDialog.tsx
@@ -114,7 +114,7 @@ export const SystemCustomLabelDialog = ({ systemId, visible, setVisible }: Syste
                   aria-describedby="username-help"
                   autoComplete="off"
                   value={label}
-                  maxLength={5}
+                  maxLength={11}
                   onChange={e => setLabel(e.target.value)}
                   // @ts-expect-error
                   ref={inputRef}
--- a/assets/js/hooks/Mapper/components/mapRootContent/components/SignatureSettings/SignatureSettings.tsx
+++ b/assets/js/hooks/Mapper/components/mapRootContent/components/SignatureSettings/SignatureSettings.tsx
@@ -11,8 +11,8 @@ import { useMapRootState } from '@/hooks/Mapper/mapRootProvider';
 import { MassState, OutCommand, SignatureGroup, SystemSignature, TimeStatus } from '@/hooks/Mapper/types';
 import { Dialog } from 'primereact/dialog';
 import { InputText } from 'primereact/inputtext';
-import { useCallback, useEffect, useState } from 'react';
-import { Controller, FormProvider, useForm } from 'react-hook-form';
+import { useCallback, useEffect, useState, inputRef} from 'react';
+import { Controller, FormProvider, useForm, useRef } from 'react-hook-form';
 
 type SystemSignaturePrepared = Omit<SystemSignature, 'linked_system'> & {
   linked_system: string;
@@ -35,9 +35,15 @@ export const SignatureSettings = ({ systemId, show, onHide, signatureData }: Map
     data: { systemSignatures, systems, wormholesData },
   } = useMapRootState();
 
-  const handleShow = async () => {};
-
   const signatureForm = useForm<Partial<SystemSignaturePrepared>>({});
+
+  const { setFocus } = signatureForm;
+
+ const handleShow = useCallback(() => {
+  setTimeout(() => {
+    setFocus("description");
+  }, 60);
+}, [setFocus]);
 
   const [userSettings, setUserSettings] = useState<any>(null);
 
@@ -266,7 +272,7 @@ export const SignatureSettings = ({ systemId, show, onHide, signatureData }: Map
                     name="description"
                     control={signatureForm.control}
                     render={({ field }) => (
-                      <InputText placeholder="Type description" value={field.value} onChange={field.onChange} />
+                      <InputText placeholder="Type description" ref ={field.ref} value={field.value} onChange={field.onChange} />
                     )}
                   />
                 </label>
--- a/assets/js/hooks/Mapper/components/mapRootContent/components/SignatureSettings/SignatureSettings.tsx
+++ b/assets/js/hooks/Mapper/components/mapRootContent/components/SignatureSettings/SignatureSettings.tsx
@@ -11,8 +11,8 @@ import { useMapRootState } from '@/hooks/Mapper/mapRootProvider';
 import { MassState, OutCommand, SignatureGroup, SystemSignature, TimeStatus } from '@/hooks/Mapper/types';
 import { Dialog } from 'primereact/dialog';
 import { InputText } from 'primereact/inputtext';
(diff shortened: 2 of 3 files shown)


SzesciennaGeometria:
screenshot for context
Zrzut ekranu 2026-07-22 221753

assets/js/hooks/Mapper/components/mapRootContent/components/SignatureSettings/SignatureSettings.tsx
```diff
@@ -11,8 +11,8 @@ import { useMapRootState } from '@/hooks/Mapper/mapRootProvider';
 import { MassState, OutCommand, SignatureGroup, SystemSignature, TimeStatus } from '@/hooks/Mapper/types';
 import { Dialog } from 'primereact/dialog';
 import { InputText } from 'primereact/inputtext';
-import { useCallback, useEffect, useState } from 'react';
-import { Controller, FormProvider, useForm } from 'react-hook-form';
+import { useCallback, useEffect, useState, inputRef} from 'react';
+import { Controller, FormProvider, useForm, useRef } from 'react-hook-form';
```
guarzo: Neither of these imports is right — `inputRef` isn't exported by `react`, and `useRef` comes from `react`, not `react-hook-form`. `tsc --noEmit` flags both. I think you want:

```ts
import { useCallback, useEffect, useState } from 'react';
import { Controller, FormProvider, useForm } from 'react-hook-form';
```

since neither `inputRef` nor `useRef` ends up being used in the final version anyway (you're going through `setFocus`).

*(edited — I'd said these "won't build", which was overstating it. The Vite build doesn't typecheck, so this may well run fine for you; it's still worth fixing, just not the emergency I made it sound like.)*

guarzo (commented):
Autofocus on the description field is a good call, that dialog is annoying without it.

You should rebase on `main`, the label input in `SystemSettingsDialog` has picked up a `disabled={hasIntelSource}` since you branched.

One general thing #630 and #631 each have a TypeScript error that a local `cd assets && npx tsc --noEmit` would catch in a few seconds. The repo runs with `strict` and `noUnusedLocals` on, so it's pretty good at catching typos before they reach a reviewer. Might save you a round trip on all three.

assets/js/hooks/Mapper/components/mapRootContent/components/SignatureSettings/SignatureSettings.tsx
```diff
@@ -35,10 +35,16 @@ export const SignatureSettings = ({ systemId, show, onHide, signatureData }: Map
...
-
   const signatureForm = useForm<Partial<SystemSignaturePrepared>>({});
 
+  const { setFocus } = signatureForm;
+
+ const handleShow = useCallback(() => {
+  setTimeout(() => {
+    setFocus("description");
+  }, 60);
+}, [setFocus]);
```
guarzo: Does this need the `setTimeout`? `onShow` already fires after PrimeReact has mounted the dialog content, so `setFocus('description')` should be able to run directly. If there is a genuine race and the timer is needed, could you pull the 60 out into a named constant with a quick comment saying what it's waiting for? Bare timing numbers tend to get flagged.

Indentation is off by a space or two here as well — `yarn format` will sort it.

assets/js/hooks/Mapper/components/mapRootContent/components/SignatureSettings/SignatureSettings.tsx
```diff
@@ -266,7 +272,7 @@ export const SignatureSettings = ({ systemId, show, onHide, signatureData }: Map
                     name="description"
                     control={signatureForm.control}
                     render={({ field }) => (
-                      <InputText placeholder="Type description" value={field.value} onChange={field.onChange} />
+                      <InputText placeholder="Type description" ref ={field.ref} value={field.value} onChange={field.onChange} />
```
guarzo: Tiny one: `ref ={field.ref}` has a stray space, should be `ref={field.ref}`. Prettier will fix it.

assets/js/hooks/Mapper/components/mapInterface/components/SystemCustomLabelDialog/SystemCustomLabelDialog.tsx
```diff
@@ -114,7 +114,7 @@ export const SystemCustomLabelDialog = ({ systemId, visible, setVisible }: Syste
                   aria-describedby="username-help"
                   autoComplete="off"
                   value={label}
-                  maxLength={5}
+                  maxLength={11}
```
guarzo, line 117: Is the `maxLength` bump related to the autofocus change, or did it come along for the ride? (Same change is in `SystemSettingsDialog.tsx`.) Not objecting to it, but it's a separate user-visible change and it'd be easier to review — and easier to revert later if it causes problems — as its own PR.

If it does stay, a screenshot would help: the custom label renders on the node itself, and 5 → 11 chars is more than double the width, so I'd want to see what a full-length label does to node layout with a long system name next to it. Also worth checking whether anything validates the length server-side, since `maxLength` only stops typing in the input, not a value arriving through the API.

SzesciennaGeometria: yes it came along just because i didnt figured out how git branching works at that point.
its so small i dont think this will break anything. i also included screenshot of this change in 1st reply to my own pull

Zrzut ekranu 2026-07-22 221753

it renders werid ? yes, but who uses everysingle one of those at the same time ? 

>(Same change is in `SystemSettingsDialog.tsx`.)

yes for some reason double-slick-system and right-click-system had limit set separately

assets/js/hooks/Mapper/components/mapRootContent/components/SignatureSettings/SignatureSettings.tsx
SzesciennaGeometria: i rememeber there was something like that and those may have been added just for the reason of not compiling without them. ill comment those out temporary just in case i break something again and need to reintroduce

assets/js/hooks/Mapper/components/mapRootContent/components/SignatureSettings/SignatureSettings.tsx
SzesciennaGeometria: i honestly dont remember why it had timeout added, that was one of the 1st things i did when i started playing around in it and could sware i piggybacked of system settings->"custom name" focus when opened. 

removed


edit: fine ,3rd 1st was checking if i can change connection thickness :P

assets/js/hooks/Mapper/components/mapRootContent/components/SignatureSettings/SignatureSettings.tsx
SzesciennaGeometria: i know: _yarn format_ 
one of those days...

fixed

PR #626, opened 2026-07-22
feat: coloring signature rows on paste

SzesciennaGeometria:
feature for old-timer and pathfinder-users: basicaly gives visual feedback that sigs got successfully pasted. (apart of refreshing timer - allowing to reduce signature window size)
"highlights" pasted signatures for 3s: greem = new, yellow = old.
should not collide with non-existent and lazy-delete

Zrzut ekranu 2026-07-22 214429


assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/helpers/rowStyles.ts
```diff
@@ -24,6 +25,15 @@ export function getSignatureRowClass(
     return clsx([...baseCls, 'bg-red-400/40 hover:bg-red-400/50']);
   }
 
+  const glowInfo = glowingRows.get(row.eve_id);
```
guarzo, line 28: `glowingRows` is declared optional on line 10 but `.get()` is called unguarded here, so with `strict` on `tsc --noEmit` flags it as possibly undefined. No runtime impact — the only caller always passes it — but worth tidying: either `glowingRows?.get(row.eve_id)`, or make the param required since `SystemSignaturesContent` defaults it.

While you're here — should the glow take priority over `row.deleted`? Right now the red lazy-delete background wins, because that check returns above this one. Probably the right call, just flagging it since the description mentions not colliding with lazy-delete and I couldn't tell if that ordering was deliberate.

*(edited — I'd opened this with "this won't compile", which was wrong: the build doesn't typecheck, so it runs fine. It's a typecheck nit, not a blocker.)*

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/hooks/useSystemSignaturesData.ts
```diff
@@ -40,6 +42,38 @@ export const useSystemSignaturesData = ({
...
+
+       const idsToRemove = incomingSignatures.map(sig => sig.eve_id);
+
+      const glowingRowsValue = settings[SETTINGS_KEYS.GLOWINGROWS_TIMING];
+      const glowingRowsIndex = glowingRowsValue && typeof glowingRowsValue === 'object' && 'value' in glowingRowsValue
+        ? (glowingRowsValue as any).value
+        : glowingRowsValue;
+
+      const glowingRowsTimeout = [0, 1000, 5000, 10000, 30000];
+      const glowingrowsTimeoutDuration = glowingRowsTimeout[glowingRowsIndex] ?? 1000;
```
guarzo: This duplicates the helper you already wrote — `getGlowingrowsTimeoutMs()` in `constants.ts:204` does exactly this (including the object-vs-primitive unwrap) and nothing calls it, so it's dead code as it stands.

Swapping these seven lines for `const glowingrowsTimeoutDuration = getGlowingrowsTimeoutMs(settings);` would kill the `[0, 1000, 5000, 10000, 30000]` array too, which is the bit I'd most want gone — it's positionally coupled to the enum, so reordering `SIGNATURES_GLOWINGROWS_TIMING` or inserting an option silently gives everyone the wrong duration. `SIGNATURE_GLOWINGROWS_TIMEOUTS` right above the helper already keys off the enum properly.

Bonus: that also gets rid of the `as any` cast on line 61.

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/hooks/useSystemSignaturesData.ts
```diff
@@ -40,6 +42,38 @@ export const useSystemSignaturesData = ({
...
+
+      setTimeout(() => {
+        setGlowingRows(current => {
+          if (!current) return new Map();
+          const updated = new Map(current);
+          idsToRemove.forEach(id => updated.delete(id));
+          return updated;
+        });
+      }, glowingrowsTimeoutDuration);
```
guarzo, line 66: Couple of things about this timer:

- It's never cleared, so if the widget unmounts (switching systems, closing the widget) during the window you get a setState on an unmounted component.
- Two pastes inside the window overlap, and since the first timer deletes by id it'll clear the glow from the second paste early if they share signatures. Re-pasting the same list twice is a pretty normal thing to do, so I think this is reachable.

Storing the timer in a ref and clearing it before setting a new one — plus clearing on unmount — would cover both. Or, since the removal is per-id anyway, stamping an expiry onto the map value and filtering on read is another way out.

Also `if (!current) return new Map();` on line 69 can't fire — `current` is state initialised to `new Map()` and only ever set to a Map — so that line can probably go.

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/hooks/useSystemSignaturesData.ts
```diff
@@ -40,6 +42,38 @@ export const useSystemSignaturesData = ({
         return;
       }
 
+      setGlowingRows(current => {
+        const newGlowing = new Map(current);
+        incomingSignatures.forEach(sig => {
+          const existing = signaturesRef.current.find(s => s.eve_id === sig.eve_id);
+          const isBrandNew = !existing || !existing.updated_at ||
+            Math.abs(new Date(existing.inserted_at).getTime() - new Date(existing.updated_at).getTime()) < 50;
```
guarzo, line 50: Want to double-check the green/yellow split here. If I'm reading it right, a signature that's already in the system but has never been edited has `inserted_at === updated_at`, so it lands under 50ms and gets marked `isBrandNew` — meaning it flashes green rather than yellow. So pasting the same unchanged list twice would show all-green both times, when the description says yellow is meant to mean "old".

Is `!existing` on its own not enough for the new/old call? That's the bit that actually says "this wasn't here before the paste", and `signaturesRef.current` is still pre-update at this point since `handleUpdateSignatures` hasn't been awaited yet.

If the `updated_at` comparison does need to stay, could the `50` become a named constant with a note on what it's guarding against? Bare timing numbers tend to get flagged here.

guarzo (commented):
I really like this one. "Did my paste actually land" is a genuine gap — squinting at the timer column to work it out is exactly the kind of thing you stop noticing you're doing until someone fixes it, and being able to shrink the signature window as a result is a nice side effect. Coming from Pathfinder this'll feel familiar to a lot of people.

The shape of it is good too: the glow state lives in the hook next to the paste that causes it, the row styling stays in `rowStyles.ts` where the rest of the row colouring already is, and making the duration configurable rather than picking a number for everyone was the right instinct. Screenshot in the description is appreciated as well — made it much easier to see what you were going for.

Comments below are all fixable stuff rather than anything structural. The one I'd start with is that the timeout logic ended up written twice — you've got a perfectly good `getGlowingrowsTimeoutMs` helper in `constants.ts` that nothing calls, and a hardcoded `[0, 1000, 5000, 10000, 30000]` array in the hook doing the same job. Deleting the array and calling the helper takes care of a few of the other notes at the same time. Beyond that there's a compile error in `rowStyles.ts`, an uncleared timer, and the settings migration that new stored settings need.

assets/js/hooks/Mapper/constants/signatures.ts
```diff
@@ -55,6 +64,7 @@ export const DEFAULT_SIGNATURE_SETTINGS: SignatureSettingsType = {
   [SETTINGS_KEYS.LAZY_DELETE_SIGNATURES]: true,
   [SETTINGS_KEYS.KEEP_LAZY_DELETE]: false,
   [SETTINGS_KEYS.DELETION_TIMING]: SIGNATURES_DELETION_TIMING.DEFAULT,
+  [SETTINGS_KEYS.GLOWINGROWS_TIMING]: SIGNATURES_GLOWINGROWS_TIMING.GLOWDEFAULT,
```
guarzo, line 67: Adding a key to `DEFAULT_SIGNATURE_SETTINGS` changes the shape of the stored settings object — it feeds `createDefaultStoredSettings()` (`mapRootProvider/helpers/createDefaultStoredSettings.ts:35`) as `signaturesWidget`. So anyone with settings already in localStorage won't have `glowingrows_timing`, and the new dropdown will read `undefined` for them.

That needs a migration under `mapRootProvider/migrations/list/` plus a bump to `STORED_SETTINGS_VERSION` in `version.ts` (currently 4). Adding a variable is a minor bump under the scheme DanSylvest described, and the migration should spread the full defaults and layer stored values on top rather than only rebuilding the one slice.

The duration itself falls back to 1s so it won't break, but the settings dropdown will render empty until they pick something.

(Small naming thing while I'm in here: `GLOWINGROWS` / `getGlowingrowsTimeoutMs` reads a bit oddly next to `DELETION_TIMING` / `getDeletionTimeoutMs`. `GLOW_TIMING` / `getGlowTimeoutMs` would match better, and the enum members could drop the `GLOW` prefix since the enum name already carries it.)

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/SystemSignaturesContent/SystemSignaturesContent.tsx
```diff
@@ -66,6 +67,7 @@ export const SystemSignaturesContent = ({
   selectable,
   onSelect,
   filterSignature,
+  glowingRows = new Map(),
```
guarzo: `glowingRows = new Map()` as a default builds a fresh Map on every render when the prop is undefined, and since it's now in the dep array on line 234 the row-class callback gets rebuilt every render along with it. Hoisting a module-level `const EMPTY_GLOWING_ROWS = new Map()` and defaulting to that keeps it stable.

Minor either way — the prop is always passed from `SystemSignatures.tsx` — but the default is what makes it look safe.

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/constants.ts
```diff
@@ -155,6 +157,18 @@ export const SIGNATURE_SETTINGS = {
         { value: SIGNATURES_DELETION_TIMING.EXTENDED, label: '30s' },
       ],
     },
+    {
+      type: SettingsTypes.dropdown,
+      key: SETTINGS_KEYS.GLOWINGROWS_TIMING,
+      name: 'Paste-Flash Timing',
+      options: [
+        { value: SIGNATURES_GLOWINGROWS_TIMING.GLOWIMMEDIATE, label: '0s' },
```
guarzo, line 166: What does the 0s option do here? For deletion timing 0s means "immediately", but for a flash it means the glow is removed in the same tick it's added, so I'd expect nothing visible at all. If that's meant to be the "off" switch it might read better labelled as such, or as a separate on/off flag.

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/hooks/useSystemSignaturesData.ts
```diff
@@ -4,7 +4,7 @@ import { Commands, ExtendedSystemSignature, SignatureKind } from '@/hooks/Mapper
 import { useCallback, useEffect, useState } from 'react';
 import useRefState from 'react-usestateref';
 
-import { SETTINGS_KEYS } from '@/hooks/Mapper/constants/signatures.ts';
+import { SETTINGS_KEYS} from '@/hooks/Mapper/constants/signatures.ts';
```
guarzo: Few spots `yarn format` would tidy up — probably quickest to just run it rather than fix them by hand:

- this line, `SETTINGS_KEYS}` is missing the space
- line 57, extra leading space on `const idsToRemove`
- lines 75-76, double blank line
- `constants.ts:1`, the import is ~148 chars, over the 120 printWidth
- `constants.ts:188`, missing blank line before the `/**` block that follows

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/helpers/rowStyles.ts
SzesciennaGeometria: literaly at this point i have no idea. 
i had like 6 iterations of this feature and last 2 were doen that if ANYONE tracked by the map pasted sigs rows would light up, but it bugged out sometimes if two ppl pasted one after another within lie 2s "new" sigs were overriten by "amber" color. also unnecessary load on the server so it got scrapped and replace by client-side-only... i hope.

lazydelete its fime.
tho i was thinking of changing lazydelete timmeout 0s to 1 or 2 seconds just to visualy show line is beeing deleted instead of just vanishing.
also i decided not to combine lazydelete and flash timouts since some ppl may prefer short flash and long lazydelete just becouse of "undo" and current lack of ctrl-z function if you miss-paste

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/hooks/useSystemSignaturesData.ts
SzesciennaGeometria: gone

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/hooks/useSystemSignaturesData.ts
SzesciennaGeometria: > Want to double-check the green/yellow split here. If I'm reading it right, a signature that's already in the system but has never been edited has `inserted_at === updated_at`, so it lands under 50ms and gets marked `isBrandNew` — meaning it flashes green rather than yellow. So pasting the same unchanged list twice would show all-green both times, when the description says yellow is meant to mean "old".

umm cant confirm it does that in testing maby im not pasting fast enough, 1st paste puts greens(new) in nexty ones colors them yellow(existing). id like to think its behaving as intended.

also im not shure it may have been something left from previous iteration for example guarding from simultaneous pastes from n+1 ppl

assets/js/hooks/Mapper/constants/signatures.ts
SzesciennaGeometria: at the time GLOWINGROWS was easier for tracking and not be coliding with any other possible "glow" on stuff

its the same "undefined" i attempted in "to_5.ts" from #629 ?

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/SystemSignaturesContent/SystemSignaturesContent.tsx
SzesciennaGeometria: you ment this ?

```
const EMPTY_GLOWING_ROWS = new Map()

export const SystemSignaturesContent = ({
  systemId,
  signatures,
  selectedSignatures,
  onSelectSignatures,
  onDeleteSelected,
  onSelectAll,
  onPaste,
  settings,
  hideLinkedSignatures,
  hasUnsupportedLanguage,
  selectable,
  onSelect,
  filterSignature,
  //glowingRows = new Map(),
}:

  return getSignatureRowClass(
      rowData as ExtendedSystemSignature,
      refVars.current.selectedSignatures || [],
      refVars.current.settings[SETTINGS_KEYS.COLOR_BY_TYPE] as boolean,
      //glowingRows,
      EMPTY_GLOWING_ROWS,
    );
  }, [/*glowingRows*/ EMPTY_GLOWING_ROWS]);
```

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/constants.ts
SzesciennaGeometria: it came straight out of of SIGNATURES_DELETION_TIMING.IMMEDIATE to be inline with existing lazy-delete dropdown

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/hooks/useSystemSignaturesData.ts
SzesciennaGeometria: i gues i idd need to start using _yarn format_ since its becoming an recurring issue 
tho lines 75-76 may not be mine dooing :P

constants.ts:1 - no idea what does that mean(goes outside the page if printed ?). fixed
constants.ts:188 - meh. fixed

assets/js/hooks/Mapper/components/mapInterface/widgets/SystemSignatures/hooks/useSystemSignaturesData.ts
SzesciennaGeometria: trying to do this i think i broke something 👎 
not clear if defective code went in with last commit: 
[github.com/wanderer-industries/wanderer/pull/626/commits/d440db5258482324e625882ca75feb5300e38010](url)

SzesciennaGeometria:
s*it happened, stuff got borked. 
unable to revert changes, this pull-request will be back when i manage to figure out what broke and how to fix it

PR #627, opened 2026-07-23
Include characters that haven't passaged in user activity API

ryanrolds:
I noticed that not all characters were being included in the `/api/map/character-activity`. It was limited to characters that had passages recorded. These changes expand the results to incldue all characters with activity, not only passages. The changes include removing some unused code and making some helpers the same shape.


ryanrolds:
I think the build is error do to the token for remote PRs being read-only. Do you need me to do anything to consider this change?

PR #629, opened 2026-07-25
feat(system): Animatyed border/outline for selected system with settings

SzesciennaGeometria:
additional optional border/outline with controlls to disable this effect or its animation.
minor adjustments to LABEL and UNSPLASHED offset 
prepared for future theme modyfications 

Zrzut ekranu 2026-07-25 100851
Zrzut ekranu 2026-07-25 100945
Zrzut ekranu 2026-07-25 101006

edit: some ppl complained "sometimes its hard to see selected system at a glance" - this fixes that regardless of theme

### DIFF
--- a/assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
+++ b/assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
@@ -134,9 +134,8 @@
 
     :global(body.animated-outline-active) & {
       --outline-puls: color-mix(in srgb, var(--selected-system-outline-puls-100) calc(var(--outline-puls-progress) * 100%), var(--selected-system-outline-puls-0));
-      outline: var(--selected-system-outline-width) dashed var(--outline-puls) !important;
-      outline-offset: var(--selected-system-outline-offset) !important;
-      box-shadow: var(--selected-system-shadow) !important;
+      outline: var(--selected-system-outline-width, 2px) var(--selected-system-outline-style, dashed) var(--outline-puls) !important;
+      outline-offset: var(--selected-system-outline-offset, 2px) !important;
       animation: animated-color-puls 1s ease-in-out infinite alternate !important;
     }
 
--- a/assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
+++ b/assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
@@ -11,6 +11,26 @@
   }
 }
 
+@property --animated-border-angle {
+  syntax: "<angle>";
+  initial-value: 0deg;
+  inherits: false;
+}
+@keyframes animated-border-rotate {
+  0% {--animated-border-angle: 0deg;}
+  100% {--animated-border-angle: 360deg;}
+}
+
+  @property --outline-puls-progress {
+  syntax: "<number>";
+  initial-value:  0;
+  inherits: false;
+}
+  @keyframes animated-color-puls {
+  0% {--outline-puls-progress: 0;}
+  100% {--outline-puls-progress: 1;}
+}
+
 .RootCustomNode {
   display: flex;
   width: var(--rf-node-width, 130px);
@@ -99,8 +119,36 @@
   }
 
   &.selected {
-    border-color: v.$pastel-pink;
-    box-shadow: 0 0 10px #9a1af1c2;
+  border-color: v.$pastel-pink;
+  border-image-width: 1px;
+  position: relative;
+  box-shadow: var(--selected-system-shadow);
+
+    :global(body.animated-border-active) & {
+      --color-primary: var(--selected-system-border-puls-0);
+      --color-secondary: var(--selected-system-border-puls-100);
+      --step: 90deg;
+      border-image: repeating-conic-gradient(from var(--animated-border-angle), var(--color-primary) 0deg var(--step), var(--color-secondary) var(--step) calc(var(--step) * 2)) 1 stretch;
+      animation: animated-border-rotate 2s linear infinite !important;
+    }
+
+    :global(body.animated-outline-active) & {
+      --outline-puls: color-mix(in srgb, var(--selected-system-outline-puls-100) calc(var(--outline-puls-progress) * 100%), var(--selected-system-outline-puls-0));
+      outline: var(--selected-system-outline-width) dashed var(--outline-puls) !important;
+      outline-offset: var(--selected-system-outline-offset) !important;
+      box-shadow: var(--selected-system-shadow) !important;
+      animation: animated-color-puls 1s ease-in-out infinite alternate !important;
+    }
+
+    :global(body.animated-border-active.animated-outline-active) & {
+      animation: animated-color-puls 1s ease-in-out infinite alternate, animated-border-rotate 2s linear infinite !important;
+    }
+    :global(body.animated-outbor-active) & {
+       animation-duration: 0s !important;
+    }
+    :global(body.animated-border-active.animated-outline-active.animated-outbor-active) & {
+       animation-duration: 0s, 0s !important;
+    }
   }
 
   &.rally {
@@ -178,7 +226,7 @@
     min-width: var(--rf-node-bookmark-min-width, 13px);
     height: var(--rf-node-bookmark-height, 22px);
     position: relative;
-    top: var(--rf-node-bookmark-top, -13px);
+    top: var(--rf-node-bookmark-top, -17px);
     border-radius: var(--rf-node-bookmark-border-radius, 5px);
     color: #ffffff;
     font-size: var(--rf-node-bookmark-font-size, 8px);
@@ -220,6 +268,7 @@
   z-index: -1;
   display: flex;
   flex-wrap: wrap;
+  margin-top: 2px;
   gap: var(--rf-node-unsplashed-gap, 2px);
   left: var(--rf-node-unsplashed-left, 2px);
 
@@ -427,7 +476,6 @@
     }
   }
 }
-
 .ShatteredIcon {
   position: relative;
   //top: -1px;
--- a/assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
+++ b/assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
@@ -3,12 +3,8 @@
 @use '@/hooks/Mapper/components/map/styles/solar-system-node' as v;
 
 @keyframes move-stripes {
-  from {
-    background-position: 0 0;
-  }
-  to {
-    background-position: 30px 0;
-  }
+  from {background-position: 0 0;}
+  to {background-position: 30px 0;}
 }
 
 @property --animated-border-angle {
@@ -21,12 +17,12 @@
   100% {--animated-border-angle: 360deg;}
 }
 
-  @property --outline-puls-progress {
+@property --outline-puls-progress {
   syntax: "<number>";
   initial-value:  0;
   inherits: false;
 }
-  @keyframes animated-color-puls {
+@keyframes animated-color-puls {
   0% {--outline-puls-progress: 0;}
   100% {--outline-puls-progress: 1;}
 }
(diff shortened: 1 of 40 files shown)


assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/MapSettingsProvider.tsx
```diff
@@ -58,7 +58,7 @@ export const MapSettingsProvider = ({ children }: WithChildren) => {
     async (prop: keyof UserSettings, value: boolean | string | Record<string, string>) => {
       const { userRemoteSettings, interfaceSettings, outCommand, setInterfaceSettings } = refVars.current;
 
-      if (UserSettingsRemoteList.includes(prop as any)) {
+      if (UserSettingsRemoteList.includes(prop as any) {
```
guarzo: Missing closing paren — `includes(prop as any) {` needs to be `includes(prop as any)) {`.

Looks like it happened in `b37a4080` ("code cleanup") — in the two commits before it this line read `includes(prop as any) && prop !== 'disable_animated_outlineborder'`, and removing that clause took the paren with it. Which also explains why your screenshots look right: they're from before this landed.

Worth a look at whether you still want that `disable_animated_outlineborder` exclusion, since it went away in the same edit.

assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/types.ts
```diff
@@ -3,6 +3,9 @@ import { InterfaceStoredSettings } from '@/hooks/Mapper/mapRootProvider/types.ts
 export enum UserSettingsRemoteProps {
   link_signature_on_splash = 'link_signature_on_splash',
   select_on_spash = 'select_on_spash',
+  show_animated_border = 'show_animated_border',
+  show_animated_outline = 'show_animated_outline',
+  disable_animated_outlineborder = 'disable_animated_outlineborder',
```
guarzo, line 8: I don't think these three belong in `UserSettingsRemoteProps` — they're interface settings, not remote ones. The checkbox list in `MapSettings/constants.ts` refers to them as `InterfaceStoredSettingsProps.*`, they're in `STORED_INTERFACE_DEFAULT_VALUES`, and they're not added to `UserSettingsRemoteList`, so `handleSettingChange` correctly routes them down the interface branch. The remote declarations aren't doing anything.

Adding them as required fields to `UserSettingsRemote` (lines 23-25) also leaves `useState<UserSettingsRemote>({ ...DEFAULT_REMOTE_SETTINGS })` on `MapSettingsProvider.tsx:43` short of three properties, since `DEFAULT_REMOTE_SETTINGS` doesn't have them. Nothing at runtime cares — the build doesn't typecheck — but `tsc --noEmit` flags it, and it'll bite whoever next touches that type.

I think deleting both blocks from this file is all it needs — the interface side already covers it.

guarzo (commented):
The motivation here is solid — "hard to spot the selected system at a glance" is something I've heard too, and doing it theme-independently is the right instinct. Pulling the selection colours out into `eve-common-variables.scss` so themes can override them later is a nice touch as well. Screenshots made this much easier to follow, thanks.

Two things before it goes in. A paren went missing on `MapSettingsProvider.tsx:61` in the last commit ("code cleanup") — your screenshots line up with the two commits before that, so it looks like a late slip rather than anything you tested around. And the three new settings are declared in both the remote and the interface settings types, but only the interface side is actually wired up. Rest is smaller stuff, inline.

assets/js/hooks/Mapper/mapRootProvider/constants.ts
```diff
@@ -15,6 +15,9 @@ import { DEFAULT_WIDGETS, STORED_VISIBLE_WIDGETS_DEFAULT } from '@/hooks/Mapper/
 export const STORED_INTERFACE_DEFAULT_VALUES: InterfaceStoredSettings = {
   isShowMenu: false,
   isShowKSpace: false,
+  show_animated_border: false,
+  show_animated_outline: false,
+  disable_animated_outlineborder: false,
```
guarzo, line 21: These three keys change the shape of the stored settings object, so this needs a migration under `mapRootProvider/migrations/list/` and a bump to `STORED_SETTINGS_VERSION` in `version.ts` (currently 4).

Without it, anyone with settings already in localStorage gets `undefined` for all three rather than `false`. It happens to degrade OK here since everything runs through `!!`, but the checkboxes will render unchecked-but-uncontrolled, and it's the kind of thing that bites harder the next time someone adds a non-boolean setting.

Adding a variable is a minor bump under the scheme DanSylvest described, and the migration wants to spread the full defaults and layer stored values on top rather than only rebuilding the `interface` slice.

Small naming thing while I'm here: every other key in `InterfaceStoredSettings` is camelCase (`isShowMenu`, `isThickConnections`, `isShowBackgroundPattern`). snake_case is the convention on the *remote* side. `isShowAnimatedBorder` etc. would sit better — and since this needs a migration anyway, renaming now is cheaper than later.

assets/js/hooks/Mapper/mapRootProvider/hooks/useMapUserSettings.ts
```diff
@@ -186,6 +186,20 @@ export const useMapUserSettings = ({ map_slug }: MapRootData, outCommand: OutCom
...
+
+      const hasAnimBorder = currentInterface.show_animated_border;
+      const hasAnimOutline = currentInterface.show_animated_outline;
+      const hasAnimOutBor = currentInterface.disable_animated_outlineborder;
+
+      document.body.classList.toggle('animated-border-active', !!hasAnimBorder);
+      document.body.classList.toggle('animated-outline-active', !!hasAnimOutline);
+      document.body.classList.toggle('animated-outbor-active', !!hasAnimOutBor);
+    }
+  }, [map_slug, mapUserSettings]);
```
guarzo, line 208: Is `document.body` the right place for these? There's an established pattern for exactly this — `MapRootContent.tsx:33` derives ``const themeClass = `${interfaceSettings.theme ?? 'default'}-theme` `` and puts it on the container as a normal className. Doing the same here would let the SCSS drop all the `:global(body....)` selectors and the `!important`s that come with them, and it'd keep the styling inside React's render rather than an effect reaching out to the DOM.

If it does stay on body, this effect has no cleanup — the three classes stay on `<body>` after the component unmounts, so anything else rendered on the page afterwards keeps the styles. Returning a cleanup that removes them would cover it.

There's also some overlap with `MapSettingsProvider.tsx:76-84`, which toggles the same classes on change. If the settings-driven effect here is the source of truth, the toggles over there might be redundant — the setting write should re-run this effect anyway.

assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/constants.ts
```diff
@@ -59,6 +59,21 @@ export const SYSTEMS_CHECKBOXES_PROPS: SettingsListItem[] = [
...
+  {
+    prop: InterfaceStoredSettingsProps.show_animated_outline,
+    label: 'Animated selected-system outline',
+    type: 'checkbox',
+  },
+   {
+    prop: InterfaceStoredSettingsProps.disable_animated_outlineborder,
+    label: 'Disable selected-system animation',
+    type: 'checkbox',
+  },
```
guarzo, line 76: Two thoughts on this third checkbox:

"Disable selected-system animation" sitting under two "Animated ..." checkboxes is a bit of a double negative — you end up ticking "Animated border" *and* "Disable animation" to get a static border, which took me a minute to work out from the CSS. Would a single "Animate" checkbox that's disabled unless one of the other two is on read more clearly? Or flipping it to "Static border/outline"?

Also `outbor` in `disable_animated_outlineborder` (and `animated-outbor-active`) is doing a lot of work as a name. `disable_animation` or similar would be easier for the next person.

Indentation on this entry is off by a space vs the two above — `yarn format` will fix it.

assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
```diff
@@ -11,6 +11,26 @@
...
+
+  @property --outline-puls-progress {
+  syntax: "<number>";
+  initial-value:  0;
+  inherits: false;
+}
+  @keyframes animated-color-puls {
+  0% {--outline-puls-progress: 0;}
+  100% {--outline-puls-progress: 1;}
+}
```
guarzo: Indentation is off in this block — `@property` and `@keyframes` are indented two spaces while their contents aren't, which reads as though they're nested inside something. The first pair on lines 14-22 is fine, so it's just these two. `initial-value:  0` on line 26 has a double space too.

Worth putting the keyframe steps on their own lines to match the rest of the file as well, but that's taste.

assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
```diff
@@ -99,8 +119,35 @@
   }
 
   &.selected {
-    border-color: v.$pastel-pink;
-    box-shadow: 0 0 10px #9a1af1c2;
+  border-color: v.$pastel-pink;
+  border-image-width: 1px;
+  position: relative;
+  box-shadow: var(--selected-system-shadow);
```
guarzo: This block lost an indent level — the four properties here are at two spaces where the rest of the rule's siblings are at four, so it reads as if `&.selected` closed early. The nested `:global(...)` blocks below kept the old depth, which makes it look stranger than it is.

While I'm in this block: there's a fair amount of `!important` in the rules below. Some of it is unavoidable if you're overriding from a `:global(body...)` selector, but if the theme-class approach from my `useMapUserSettings` comment works out, most of it should be able to go.

assets/js/hooks/Mapper/components/map/styles/eve-common-variables.scss
```diff
@@ -121,4 +121,12 @@ $homeDark30: color.adjust($homeBase, $lightness: -30%);
   --conn-bridge: rgba(135, 185, 93, 0.85);
   --conn-save: rgba(155, 102, 45, 0.85);
   --selected-item-bg: rgba(98, 98, 98, 0.33);
+
+  --selected-system-shadow: 0 0 10px #9a1af1c2;
+  //--selected-system-border: ;
```
guarzo: `//--selected-system-border: ;` and `//--selected-system-outline: ;` on line 129 look like leftovers — worth dropping if they're not coming back.

Related: `--selected-system-outline-width`, `-style` and `-offset` get referenced over in `SolarSystemNodeDefault.module.scss:137-138` (with fallbacks) but aren't defined here alongside the others. Intentional, or did they just not make it across? Defining them here would make the set easier to override from a theme, which seems to be the point of putting them in this file.

assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
```diff
@@ -178,7 +225,7 @@
     min-width: var(--rf-node-bookmark-min-width, 13px);
     height: var(--rf-node-bookmark-height, 22px);
     position: relative;
-    top: var(--rf-node-bookmark-top, -13px);
+    top: var(--rf-node-bookmark-top, -17px);
```
guarzo, line 224: This one and the `margin-top: 2px` on the unsplashed container further down — the bookmark `top` going -13px → -17px (plus -14px → -18px in `accessible-large-theme.scss`) looks unrelated to the animated border.

They're also a visual change for *everyone*, not just people who turn the new setting on, which makes them a bit surprising to find in a PR about selection styling. Are they needed to stop the outline overlapping the bookmarks? If so, that's worth a comment in the code saying so. If they're a separate fix you spotted along the way, they'd be easier to review — and easier to revert if someone disagrees about the spacing — as their own PR.

Either way a before/after on a node with bookmarks would help, since the screenshots show the border rather than that area.

assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/MapSettingsProvider.tsx
SzesciennaGeometria: parenthseses readded, but i think i swapped adisable_animated_border/outline for a single "disable-animation" for both together (less checkboxes) - it had few iterations some orphaned code may apply...

assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/types.ts
SzesciennaGeometria: i belive you found few lines like that already.
i lack grasp what should go where so when i add something and se "select_on_spash = 'select_on_spash',"
i add mine new one next to it since that "may" be important.

assets/js/hooks/Mapper/mapRootProvider/constants.ts
SzesciennaGeometria: i dont know if im getig this right, but that has to do with loading interface after new functions were added ? 
i conjured "to_5.ts" but im slightly geting out of my depth, need more info.

edit: standby for next commit

assets/js/hooks/Mapper/mapRootProvider/hooks/useMapUserSettings.ts
SzesciennaGeometria: i just assumed this can go here with all other "useEffect(() => {"
should this be moved ? i didnt throw any erros at me and "worked"

assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/constants.ts
SzesciennaGeometria: i figured that someone may like to use "only" animated border OR animated outline, there is possible to use BOTH at the same time. 
but on the other hand someone may not like ANIMATION at all. there was an iteration of this with two separated "disable border animation" and "disable outline animation" but i just merged then together since this looked confusing in the settings. 
i named "animated-outbor" and "animated_outlineborder" just in case there was some other "animation" going on at some point

is it possible to make checkbox "apear" if something is selected ?

assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
SzesciennaGeometria: fixed. i think, i use "phoenix code" it may have setting to show that i never went throu all te setting, tho i know its missing few fun features compared to ie "idea" but for some reason i cant get myself to start working that one 
Zrzut ekranu 2026-08-13 124719

assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
SzesciennaGeometria: obraz

assets/js/hooks/Mapper/components/map/styles/eve-common-variables.scss
SzesciennaGeometria: yes those may have been leftovers since i had to cut up my for to smaller peaces to separate features.
those lines are absent in #631

assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
SzesciennaGeometria: yes, it moves those labels a bit up so text wont collide with outline when enabled i figued that moving them 4 pixels wont do much of a visual difference and achived "intended" effect without any additional code

Zrzut ekranu 2026-08-13 131342
Zrzut ekranu 2026-08-13 131255

i havent seen many comment lest in the code should i start putting those more when adding/changing stuff for review ?

SzesciennaGeometria:
it was brought to my attention that (my gues is since last update) those animated borders  may not be showing on chrome for some reason. 
chrome Version 151.0.7922.137  - works
chrome Version 151.0.7922.175 - borked
i will look into that soon-ish...

edit: works fine on my testrig-rig and prod Version 152.0.7977.64 soooo i gues update and problem solved.
your mileage may vary....

SzesciennaGeometria:
i messed something up while syncing fork, i hope it rolled back to proper checkpoint

PR #630, opened 2026-07-25
feat: set any connection as bridge

SzesciennaGeometria:
yeah, pathfinder had it, and i think some asked for this on discord
its just for ease-of-navigation so can be toggled at will so ppl can see where to go - those who will need it will use it. 

Zrzut ekranu 2026-07-25 122730

i gues could add some "if neighboring k-space-systems then set-gate", but this wont be used that much to bother faster do disconnect and draw new line if someone missclicks.

### DIFF
--- a/assets/js/hooks/Mapper/components/map/components/ContextMenuConnection/ContextMenuConnection.tsx
+++ b/assets/js/hooks/Mapper/components/map/components/ContextMenuConnection/ContextMenuConnection.tsx
@@ -20,7 +20,7 @@ export interface ContextMenuConnectionProps {
   onChangeShipSizeStatus(state: ShipSizeStatus): void;
   onChangeType(type: ConnectionType): void;
   onToggleMassSave(isLocked: boolean): void;
-  onToggleBbridge(isBridge: boolean): void;
+  onToggleBridge(isBridge: boolean): void;
   onHide(): void;
   edge?: Edge<SolarSystemConnection>;
 }
@@ -67,11 +67,6 @@ export const ContextMenuConnection: React.FC<ContextMenuConnectionProps> = ({
 
     if (edge.data?.type === ConnectionType.gate) {
       return [
-         {
-          label: `Set as Bridge`,
-          icon: 'pi hero-forward',
-          command: () => onChangeType(ConnectionType.bridge),
-        },
         {
           label: 'Disconnect',
           icon: PrimeIcons.TRASH,
@@ -124,12 +119,7 @@ export const ContextMenuConnection: React.FC<ContextMenuConnectionProps> = ({
         command: () => onToggleMassSave(!edge.data?.locked),
       },
       ...(bothNullsec
-        ? [/*
-            { //ability only ros bridge between nullsecs
-              label: `Set as Bridge`,
-              icon: 'pi hero-forward',
-              command: () => onChangeType(ConnectionType.bridge),
-            },*/
+        ? [
           ]
         : []),
       {
--- a/assets/js/hooks/Mapper/components/map/components/ContextMenuConnection/ContextMenuConnection.tsx
+++ b/assets/js/hooks/Mapper/components/map/components/ContextMenuConnection/ContextMenuConnection.tsx
@@ -20,6 +20,7 @@ export interface ContextMenuConnectionProps {
   onChangeShipSizeStatus(state: ShipSizeStatus): void;
   onChangeType(type: ConnectionType): void;
   onToggleMassSave(isLocked: boolean): void;
+  onToggleBbridge(isBridge: boolean): void;
   onHide(): void;
   edge?: Edge<SolarSystemConnection>;
 }
@@ -32,6 +33,7 @@ export const ContextMenuConnection: React.FC<ContextMenuConnectionProps> = ({
   onChangeShipSizeStatus,
   onChangeType,
   onToggleMassSave,
+  onToggleBridge,
   onHide,
   edge,
 }) => {
@@ -65,6 +67,11 @@ export const ContextMenuConnection: React.FC<ContextMenuConnectionProps> = ({
 
     if (edge.data?.type === ConnectionType.gate) {
       return [
+         {
+          label: `Set as Bridge`,
+          icon: 'pi hero-forward',
+          command: () => onChangeType(ConnectionType.bridge),
+        },
         {
           label: 'Disconnect',
           icon: PrimeIcons.TRASH,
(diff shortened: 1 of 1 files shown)


assets/js/hooks/Mapper/components/map/components/ContextMenuConnection/ContextMenuConnection.tsx
```diff
@@ -20,6 +20,7 @@ export interface ContextMenuConnectionProps {
   onChangeShipSizeStatus(state: ShipSizeStatus): void;
   onChangeType(type: ConnectionType): void;
   onToggleMassSave(isLocked: boolean): void;
+  onToggleBbridge(isBridge: boolean): void;
```
guarzo: Typo here — the interface says `onToggleBbridge` (double b) but line 36 destructures `onToggleBridge`. `tsc --noEmit` flags it; at runtime you just get `undefined`, which is why it doesn't show up when you use it.

Worth a second look while you're in there: `onToggleBridge` doesn't actually get called anywhere in the component, it's only listed in the `useMemo` deps on line 148. The new menu items call `onChangeType(ConnectionType.bridge)` instead. So either the prop isn't needed at all and both it and the dep can go, or the menu items were meant to call it. Whichever you intended, right now it's half-wired.

*(edited — I'd said this "won't compile". Not true, sorry: `vite build` doesn't typecheck, so nothing here would have stopped it running.)*

assets/js/hooks/Mapper/components/map/components/ContextMenuConnection/ContextMenuConnection.tsx
```diff
@@ -112,12 +124,12 @@ export const ContextMenuConnection: React.FC<ContextMenuConnectionProps> = ({
...
       ...(bothNullsec
-        ? [
-            {
+        ? [/*
+            { //ability only ros bridge between nullsecs
               label: `Set as Bridge`,
               icon: 'pi hero-forward',
               command: () => onChangeType(ConnectionType.bridge),
-            },
+            },*/
```
guarzo: Could this whole block come out? With the item commented out, `...(bothNullsec ? [ ] : [])` spreads an empty array either way, so it's dead — and `bothNullsec` above it becomes pointless too.

Commented-out code tends to get flagged in review here, and git remembers the old version anyway if anyone wants it back.

assets/js/hooks/Mapper/components/map/components/ContextMenuConnection/ContextMenuConnection.tsx
```diff
@@ -65,6 +67,11 @@ export const ContextMenuConnection: React.FC<ContextMenuConnectionProps> = ({
 
     if (edge.data?.type === ConnectionType.gate) {
       return [
+         {
+          label: `Set as Bridge`,
+          icon: 'pi hero-forward',
+          command: () => onChangeType(ConnectionType.bridge),
+        },
```
guarzo: (and the near-identical block further down, around line 113)

Same menu item is now defined twice, once in the gate branch and once in the general list. Could it be pulled out into a single `const bridgeItem = { ... }` above the branches and spread into both? Saves them drifting apart later when someone changes the icon or the label.

Small consistency thing: elsewhere in the app this connection type is labelled "Jumpgate" (`components/map/constants.ts:718`), so "Set as Bridge" here is a slightly different name for the same thing. Worth settling on one name, and it might also answer the question in my top-level comment about what this type is meant to represent.

Indentation on both new blocks is off by a space — `yarn format` will handle it.

One behaviour question: with both insertions, "Set as Bridge" now shows on wormhole connections as well as gates. Is that deliberate, or a side effect of adding it in two places?

guarzo (commented):
Could you say a bit in the description about what you're using this for? The nullsec-only restriction was presumably there for a reason, and knowing the case you're covering would make it much easier for someone with merge rights to judge whether widening it is the right call or whether it wants a separate connection type.

Two things on the code itself: there's a typo in the props interface that stops it compiling, and the old nullsec branch is left in as a commented-out block that I think can just go. Notes inline.

Would you mind adding a screenshot of the menu for a wormhole and for a gate? Mostly so it's clear where the new entry sits relative to Disconnect.

assets/js/hooks/Mapper/components/map/components/ContextMenuConnection/ContextMenuConnection.tsx
SzesciennaGeometria: o i remeber this one, yes it could go out, but that was fragment of original code and if i remember right i wanted to keep i in for reference. 

ooorrr, thera was a segment of code that was introducing "duplicate" option for set as brige (there were 2 when clicked on ns connection only)

SzesciennaGeometria:
> Could you say a bit in the description about what you're using this for? The nullsec-only restriction was presumably there for a reason

> Would you mind adding a screenshot of the menu for a wormhole and for a gate? Mostly so it's clear where the new entry sits relative to Disconnect.

jump bridges can be set up onlu in null and it makes sense to be only to mark them between nullsec systems, but this feature was used in the pathfinder (and i belive we used that also in siggy) for ease-of-navigation so can be toggled to show ppl suggested route between separate wh-chains. 

requested screenshots of the connection menu
Zrzut ekranu 2026-08-11 202935
Zrzut ekranu 2026-08-11 202948
Zrzut ekranu 2026-08-11 203144

as for usecase, we used "set as bridge" like this just to sho "whitch way to go".
Zrzut ekranu 2026-08-11 203602

it was just simple for me to remove limitations on setting bridge, than introducing new connection type just for the sake of it. im shure this will be used only by those who used it in that way before.

assets/js/hooks/Mapper/components/map/components/ContextMenuConnection/ContextMenuConnection.tsx
SzesciennaGeometria: this landed in there becouse it coudnt be set as bridge for two neighboring systems. 
in hindsight yes this is pointless to work like that and ill remove it :P

> One behaviour question: with both insertions, "Set as Bridge" now shows on wormhole connections as well as gates. Is that deliberate, or a side effect of adding it in two places?

yes that was intentional for its usecase since connecting two far-away k-space ystems connects them as "wormhole" (as intended i gues)

edit: im not shure i touched that "jumpgate" thingy

assets/js/hooks/Mapper/components/map/components/ContextMenuConnection/ContextMenuConnection.tsx
SzesciennaGeometria: nice cathc, "it worked" so it went pass me. 
im lacking grasp on what goes where so when i saw "onToggleMassSave(isLocked: boolean): void;" so first thing i just added my new one next to it. i belive you allredy found few more like this.

DanSylvest:
Ok. That PR can not be passed. And i will close it. 
I understood your point. You can make feature-request and i will try to implement later. Or do it as i wrote above

PR #631, opened 2026-07-28
fix: Future-Proofing THEME creation + new FANABERIA theme

SzesciennaGeometria:
hello. finished up with my branch rewrite to be compatible with MAIN-repo.
-on the way i rewrote few variables so new themes should be easier to create just by tinkering with -theme.scss.
-did one override for custom system name - could figure that one out any other way
-added additional middle-layer to connection having same color as the "normal" top.layer - was planning to use it as -background -for dashed-mass.
-also new theme \o/

Zrzut ekranu 2026-07-28 222808
Zrzut ekranu 2026-07-28 223807

### DIFF
--- a/assets/js/hooks/Mapper/components/map/components/SolarSystemEdge/SolarSystemEdge.module.scss
+++ b/assets/js/hooks/Mapper/components/map/components/SolarSystemEdge/SolarSystemEdge.module.scss
@@ -20,10 +20,10 @@
   }
 
   &.TimeCrit {
-    stroke: var(--rf-edge-time-critical, #f11ab2);
-    stroke-width: var(--rf-edge-time-critical-width, 4px);
-    stroke-dasharray: var(--rf-edge-time-critical-dasharray, none);
-    stroke-linecap: var(--rf-edge-time-critical-linecap, butt);
+    stroke: var(--rf-edge-time-supercritical, #f11ab2);
+    stroke-width: var(--rf-edge-time-supercritical-width, 4px);
+    stroke-dasharray: var(--rf-edge-time-supercritical-dasharray, none);
+    stroke-linecap: var(--rf-edge-time-supercritical-linecap, butt);
   }
 
   &.Hovered {
@@ -38,7 +38,7 @@
     }
 
     &.TimeCrit {
-      stroke: var(--rf-edge-time-critical-hover, #ef7dce);
+      stroke: var(--rf-edge-time-supercritical-hover, #ef7dce);
     }
   }
 
@@ -54,7 +54,7 @@
     }
 
     &.TimeCrit {
-      stroke-width: var(--rf-edge-time-critical-thick-width, 6px);
+      stroke-width: var(--rf-edge-time-supercritical-thick-width, 6px);
     }
   }
 
--- a/assets/js/hooks/Mapper/components/map/components/SolarSystemEdge/SolarSystemEdge.module.scss
+++ b/assets/js/hooks/Mapper/components/map/components/SolarSystemEdge/SolarSystemEdge.module.scss
@@ -70,6 +70,12 @@
   }
 }
 
+.EdgePathMiddle {
+  fill: none;
+  stroke: var(--rf-edge-mass-normal, #2c3844);
+  stroke-width: var(--rf-edge-mass-normal-width, 2px);
+}
+
 .EdgePathFront {
   fill: none;
   stroke: var(--rf-edge-mass-normal, #2c3844);
@@ -78,12 +84,14 @@
   &.MassVerge:not(&.Frigate) {
     stroke: var(--rf-edge-mass-verge, #af0000);
     stroke-dasharray: var(--rf-edge-mass-verge-dasharray, none);
+    stroke-dashoffset: var(--rf-edge-mass-verge-dasshofset, 0);
     stroke-linecap: var(--rf-edge-mass-verge-linecap, butt);
   }
 
   &.MassHalf:not(&.Frigate) {
     stroke: var(--rf-edge-mass-half, #ffd700);
     stroke-dasharray: var(--rf-edge-mass-half-dasharray, none);
+    stroke-dashoffset: var(--rf-edge-mass-half-dasshofset, 0);
     stroke-linecap: var(--rf-edge-mass-half-linecap, butt);
   }
 
--- a/assets/js/hooks/Mapper/components/map/components/SolarSystemEdge/SolarSystemEdge.module.scss
+++ b/assets/js/hooks/Mapper/components/map/components/SolarSystemEdge/SolarSystemEdge.module.scss
@@ -84,14 +84,14 @@
   &.MassVerge:not(&.Frigate) {
     stroke: var(--rf-edge-mass-verge, #af0000);
     stroke-dasharray: var(--rf-edge-mass-verge-dasharray, none);
-    stroke-dashoffset: var(--rf-edge-mass-verge-dasshofset, 0);
+    stroke-dashoffset: var(--rf-edge-mass-verge-dashoffset, 0);
     stroke-linecap: var(--rf-edge-mass-verge-linecap, butt);
   }
 
   &.MassHalf:not(&.Frigate) {
     stroke: var(--rf-edge-mass-half, #ffd700);
     stroke-dasharray: var(--rf-edge-mass-half-dasharray, none);
-    stroke-dashoffset: var(--rf-edge-mass-half-dasshofset, 0);
+    stroke-dashoffset: var(--rf-edge-mass-half-dashoffset, 0);
     stroke-linecap: var(--rf-edge-mass-half-linecap, butt);
   }
 
--- a/assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
+++ b/assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
@@ -100,7 +100,8 @@
 
   &.selected {
     border-color: v.$pastel-pink;
-    box-shadow: 0 0 10px #9a1af1c2;
+    //refan //box-shadow: 0 0 10px #9a1af1c2;
+    box-shadow: var(--selected-system-shadow);
   }
 
   &.rally {
@@ -114,6 +115,7 @@
       z-index: -1;
 
       border-color: v.$neon-color-1;
+    /* //refan
       background: repeating-linear-gradient(
           45deg,
           v.$neon-color-3 0px,
@@ -121,6 +123,13 @@
           transparent 8px,
           transparent 21px
       );
+      */
+      background: var(--eve-solar-system-status-rally, repeating-linear-gradient(
+          45deg,
+          v.$neon-color-3 0px,
+          v.$neon-color-3 8px,
+          transparent 8px,
+          transparent 21px));
       background-size: 30px 30px;
       animation: move-stripes 3s linear infinite;
     }
@@ -128,7 +137,8 @@
 
   &.eve-system-status-home {
     border: 1px solid var(--eve-solar-system-status-color-home-dark30);
-    background-image: linear-gradient(45deg, var(--eve-solar-system-status-color-background), transparent);
+    //refan //background-image: linear-gradient(45deg, var(--eve-solar-system-status-color-background), transparent);
+    background-image: var(--eve-solar-system-status-home);
     &.selected {
       border-color: var(--eve-solar-system-status-color-home);
     }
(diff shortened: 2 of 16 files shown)


assets/js/hooks/Mapper/components/map/components/SolarSystemEdge/SolarSystemEdge.module.scss
```diff
@@ -20,10 +20,10 @@
...
 
   &.TimeCrit {
-    stroke: var(--rf-edge-time-critical, #f11ab2);
-    stroke-width: var(--rf-edge-time-critical-width, 4px);
-    stroke-dasharray: var(--rf-edge-time-critical-dasharray, none);
-    stroke-linecap: var(--rf-edge-time-critical-linecap, butt);
+    stroke: var(--rf-edge-time-supercritical, #f11ab2);
+    stroke-width: var(--rf-edge-time-supercritical-width, 4px);
+    stroke-dasharray: var(--rf-edge-time-supercritical-dasharray, none);
+    stroke-linecap: var(--rf-edge-time-supercritical-linecap, butt);
```
guarzo, line 26: How is `.TimeCrit` meant to get applied? As far as I can find it never is — `SolarSystemEdge.tsx:91-92` only ever sets `classes.time1` and `classes.time4` off `TimeStatus`, and `TimeStatus` (`types/connection.ts`) goes `_1h, _4h, _4h30m, _16h, _24h, _48h` with nothing below an hour. Grepping for `TimeCrit` across the `.ts`/`.tsx` files turns up nothing at all, so the class looks like it's been dead in the tree for a while.

If that's right, then everything hung off it here — and the `--rf-edge-time-supercritical-*` values in both new themes and the two accessible themes — has no effect yet, which would make this hard for anyone to review by looking at it.

No objection to the idea itself: supercritical is a real state and it's reasonable to want it styled separately from the 1h crit. It just seems like it needs a `TimeStatus` value and something in `SolarSystemEdge.tsx` setting the class before the variables mean anything. Is that meant to be a follow-up? Might be worth saying so in the description, or splitting the supercritical part out so the rest of the refactor can go in on its own.

assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
```diff
@@ -114,59 +117,62 @@
...
   }
 
   &.eve-system-status-target {
-    background-image: linear-gradient(275deg, var(--eve-solar-system-status-target), transparent);
+    background-image: var(--eve-solar-system-status-target);
   }
 }
 
+.RootCustomNode .BottomRow * { //node custom name override
+  font-size: var(--rf-custm-name-font-size);
```
guarzo: Typo — `--rf-custm-name-font-size` is missing the `o` in "custom". Nothing defines that name, and there's no fallback on this one, so the `font-size` here is a no-op.

Slightly bigger thing about this rule generally: it isn't scoped to the new themes, so `color: var(--rf-custom-name, #d6d3d1)` on line 173 applies to every theme, and only the two Fanaberia files define `--rf-custom-name`. That means the bottom row changes colour in default/pathfinder/accessible too. Is that intended? If not, scoping the block under the theme classes (or giving the fallback the current default colour) would keep the other themes where they are.

guarzo (commented):
Moving the hardcoded gradients and shadows out into variables is a good direction — that's the part that makes new themes possible without touching the node styles, and it looks right to me.

The one thing I'd want to understand before this goes in is the supercritical edge work. Critical and supercritical are both real states in game, so having variables for both makes sense — but as far as I can tell the selector they're hooked up to never gets applied, so none of it renders today. Details inline, and I might well be missing where it comes from.

Screenshots would help a lot here too, since almost all of this is visual — ideally both new themes, plus one of an existing theme (High-contrast especially) so it's clear nothing regressed there.

assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeTheme.module.scss
```diff
@@ -4,7 +4,7 @@
    Only override what's different from the base
    Currently none required
 ---------------------------------------------- */
-
+/* //commenting thios out, its annoing to troubleshoot when active since overriding HOME system when active
```
guarzo, line 7: Reading the note, it sounds like this is off while you troubleshoot rather than permanently — in which case it probably wants to come back before merge?

If it turns out it genuinely is redundant now that the base module reads `--eve-solar-system-status-*`, could it just be deleted instead of commented out? Same effect, and commented-out blocks tend to get flagged here. Git will still have it if it's needed later.

assets/js/hooks/Mapper/components/map/styles/fanaberia-theme.scss
```diff
@@ -0,0 +1,100 @@
+//yo mamma so fat when she takes the gate she can be seen on both locals at the same time
+//and when Obi-Wan Kenobi saw her, he said: "That's no moon..."
```
guarzo: Any chance these could come out? Same two lines are at the top of `fanaberiadashed-theme.scss`. Made me laugh, but upstream files tend to get read by people who have no context for them.

assets/js/hooks/Mapper/components/map/styles/fanaberia-theme.scss
```diff
@@ -0,0 +1,100 @@
...
+  --eve-solar-system-status-lookingFor: linear-gradient(330deg,#001c30ff, #006aefff, #001c30ff, #006aefff, #001c30ff);
+  --eve-solar-system-status-color-lookingFor: #006aefff;
+  --eve-solar-system-status-warning: linear-gradient(330deg, #001c30ff, #b0a312, #001c30ff, #b0a312, #001c30ff, #b0a312, #001c30ff);
+  --eve-solar-system-status-color-warning: #b0a312;
+  --eve-solar-system-status-target: radial-gradient(circle, #001c30ff, #7500bd, #001c30ff, #7500bd, #001c30ff, #7500bd, #001c30ff, #7500bd);
+  --eve-solar-system-status-color-target: #7500bd;
+  --eve-solar-system-status-dangerous: linear-gradient(330deg,#001c30ff,  #c75919, #001c30ff,  #c75919, #001c30ff, #c75919, #001c30ff, #c75919, #001c30ff);
+  --eve-solar-system-status-color-dangerous:  #c75919;
+  --eve-solar-system-status-rally: repeating-linear-gradient(45deg, #3182d6 0px,#3182d6 8px,transparent 8px,transparent 21px);
+  --eve-solar0system-status-color-rally: #3182d6;
```
guarzo: Small typo — `--eve-solar0system-status-color-rally` has a `0` where the hyphen should be, so this one isn't doing anything. Same line in `fanaberiadashed-theme.scss:48`.

assets/js/hooks/Mapper/components/map/styles/fanaberia-theme.scss
```diff
@@ -0,0 +1,100 @@
...
+  --rf-edge-time-normal-width: 7px;
+  --rf-edge-time-normal-thick-width: 12px;
+  --rf-edge-time-eol-width: 7px;
+  --rf-edge-time-eol-thick-width: 12px;
+  --rf-edge-time-eol-linecap: butt;
+  --rf-edge-time-critical-width: 7px;
+  --rf-edge-time-critical-thick-width: 12px;
+  --rf-edge-time-critical-linecap: butt;
+  --rf-edge-time-supercritical-width: 7px;
+  --rf-edge-time-supercritical-thick-width: 4px;
```
guarzo: Is `--rf-edge-time-supercritical-thick-width: 4px` right? It's thinner than the 7px non-thick width on the line above, and the other three pairs in this file all go the other way (7 → 12), so I'd guess it was meant to be 12. Same values in `fanaberiadashed-theme.scss:62`.

(Only matters once the supercritical class is actually applied — see my other comment.)

assets/js/hooks/Mapper/components/map/components/SolarSystemEdge/SolarSystemEdge.module.scss
```diff
@@ -78,12 +84,14 @@
   &.MassVerge:not(&.Frigate) {
     stroke: var(--rf-edge-mass-verge, #af0000);
     stroke-dasharray: var(--rf-edge-mass-verge-dasharray, none);
+    stroke-dashoffset: var(--rf-edge-mass-verge-dasshofset, 0);
```
guarzo: `dasshofset` should be `dashoffset` — same on line 94, and in both theme files (`fanaberia-theme.scss:66,69`). It's spelled the same way in all of them so it does work today, which is exactly why it's worth fixing now rather than after someone copies it into a third theme.

assets/js/hooks/Mapper/components/map/styles/fanaberiadashed-theme.scss
```diff
@@ -0,0 +1,106 @@
+//yo mamma so fat when she takes the gate she can be seen on both locals at the same time
+//and when Obi-Wan Kenobi saw her, he said: "That's no moon..."
+
```
guarzo: This file looks like a near-copy of `fanaberia-theme.scss` with a handful of edge values changed. Would it work to have the dashed one `@use` the base and override just the differing variables? Otherwise every future tweak has to be made twice — and the two typos above are already in both copies, which is sort of the point.

assets/js/hooks/Mapper/components/map/components/SolarSystemEdge/SolarSystemEdge.module.scss
SzesciennaGeometria: there is no visual fo SUPERCRITICAL avaluble (but in my theme its double the dashes from 1h-eol (critical)) 
at some point going throu the code i stumbled on commented out section where lifetime <0 was defined(?)/mentioned(?) "something something disabled, breaks something"

soo i didnt touched that, just renamed those variablwes to be more telling:

this is oryginal "closure immminent" i gues
```
.TimeCrit {  
    stroke: var(--rf-edge-time-critical, #f11ab2);
```
and this is 1h eol
```
.time1 { 
    stroke: var(--rf-edge-time-critical,
```
i just wanted to prepare theme variables for when this feature goes live

assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeDefault.module.scss
SzesciennaGeometria: yes, i introduced this variable with no fallback since it had none before i figured it can use "default state" defined elswhere.

assets/js/hooks/Mapper/components/map/components/SolarSystemNode/SolarSystemNodeTheme.module.scss
SzesciennaGeometria: actualy i didnt plan to uncomment it, but left the code intact on purpouse.
if i remember right for some reason it overrides other "home" theme setings atleast it did mine. not shure why but i spent few hours figuring whats going on why i cant chage color, and that was the culprit

assets/js/hooks/Mapper/components/map/styles/fanaberia-theme.scss
SzesciennaGeometria: fine, next time ill throw in more coder-frendly joke :P

assets/js/hooks/Mapper/components/map/styles/fanaberia-theme.scss
SzesciennaGeometria: yeah and i belive this goes nowhere anyway, not shure where else this variable should be used but i notieced there are few places that dont react to those variables as they should. not important atm. fixed

assets/js/hooks/Mapper/components/map/styles/fanaberia-theme.scss
SzesciennaGeometria: correct, went undisturbed by QA since its not used atm

assets/js/hooks/Mapper/components/map/components/SolarSystemEdge/SolarSystemEdge.module.scss
SzesciennaGeometria: i need better editing software to catch those typos... fixed

assets/js/hooks/Mapper/components/map/styles/fanaberiadashed-theme.scss
SzesciennaGeometria: @use like in "high contrast / large" i thought to do it, but i went with plain cop-paste since "dashed" verssion was ment to be more friendly for colorblind, and since i never done that kind of support befeore i planned to receive somekind of feedback and inadvertly change stuff out.

PR #632, opened 2026-07-30
Configurable labels, dangerous and bubbled connections, undo

z4nr34l:
Additions around system labels and marking connections. They are independent commits.

Export/import has moved out to #656 as @DmitryPopov asked.

### Configurable system labels

Labels on the system context menu were six hardcoded ids with fixed names and colours (`Label A`, `Label B`, ...), and auto tag / auto label on jump could only pick one of four preset index schemes.

- Labels are defined per user per map in a new **Labels** tab of the map user settings: name, colour, order, add and remove. Stored under a new `system_labels` key in map user settings.
- The name is used both in the context menu and as the badge on the system, so `WH-1` shows up as typed.
- Label **ids** stay shared map data - only their presentation is per user. Ids without a definition (someone removed a label that is still on a system) still render, with the raw id and a neutral colour, so nothing silently disappears from a map.
- Auto label, auto tag and the wormhole temporary name accept the bookmark format variables instead of a fixed dropdown, so a chain bookmarked `WH-{chain_index}` can carry the same value onto the systems. Previously stored dropdown values (`index`, `chain_index`, `index_letter`, `chain_index_letters`) map onto the equivalent template, so existing settings keep working.
- Defaults reproduce the current look (`A B C 1 2 3` with the existing colours), so an untouched map renders exactly as before.

### Dangerous and bubbled connections

A camped jumpgate is a property of the gate, not of the systems it links, and a bubble could not be recorded at all - which is the thing most likely to make you route around a hole.

- Connections carry `dangerous` (boolean) and `bubbled` (0 none, 1 source, 2 target, 3 both), set from the connection context menu.
- A dangerous connection is drawn in the dangerous colour, and the Ansiblex `B` badge turns red; connections without a badge of their own get a warning badge, so the mark is not only in the line colour.
- A bubbled end is drawn as a ring around the connection dot, in the edge layer, so the system card sits over it.
- Routes settings gained **Avoid dangerous bridges** and **Avoid bubbled connections**. Both off by default, and a client that does not send them keeps the current behaviour. Dangerous filtering applies to bridges; bubbled applies to any connection type, since bubbles sit on gates and holes too.

### Bubbled signatures

Whether a hole is bubbled matters before jumping it - by the time the connection can be marked, someone has already gone through to find out. Wormhole signatures carry an `isBubbled` flag in their existing custom info (no migration), set from the signature editor. The list shows it beside the EOL and mass markers and tints the row amber so it reads while scanning.

### Styling the bubble

@SzesciennaGeometria asked for variable names to theme the rings with. Every part of the bubble now comes from a CSS variable, so a theme can restyle it without touching the component:

| variable | what it sets | default |
| --- | --- | --- |
| `--rf-edge-bubble` | ring colour | `#ffb03a` |
| `--rf-edge-bubble-fill` | fill inside the ring | same colour at 18% |
| `--rf-edge-bubble-size` | radius (a geometry property, so CSS can set it) | `24px` |
| `--rf-edge-bubble-border` | ring thickness | `2px` |
| `--rf-edge-bubble-dash` | dash pattern | `6 4` |

For people who would rather not write a theme, the Connections tab of the map user settings exposes colour, size, border and fill. Those are stored server-side with the rest of the map user settings, so they follow the user between browsers and are carried into the map defaults an admin saves. A field left empty emits no variable at all, so the theme stays in charge - it is an override, not a replacement.

### Undo a deletion

Removing the wrong system or connection meant rebuilding it by hand while the chain carried on around it.

- Deletions are snapshotted before they reach the server - a system together with every connection that touched it - and `ctrl+z` replays the last one, so systems come back at their old coordinates with their name, description, labels, status, tag and temporary name.
- Every delete path funnels through `useDeleteSystems`, so the Delete key, the system context menu and multi select all record. Connections record from their own context menu.
- The stack holds the last twenty deletions for the session; the hotkey stays out of inputs and out of other windows.
- It is per session and per user: it cannot take back someone else's delete, and it covers deletions only, not edits.

Restoring goes through the paste path, which is why `fix(map): keep system attributes when pasting into another map` is in this branch. Pasting a system that the target map did not already have created it from static info alone and dropped its status, tag, labels, description and temporary name - the case copy/paste and undo both depend on. Covered by an integration test.

Widgets also got a blurred backdrop, so the map behind them reads as context rather than noise.

### Schema check on container start

The image starts the release directly, so a deploy that skipped `db migrate` boots against a database missing whatever the new version added - including the migration in this branch - and the failures that follow read as application bugs rather than a missed step.

The entrypoint now checks for pending migrations before starting and applies them, reusing the same interweave ordering as the manual task. Nothing pending means no work. `WANDERER_AUTO_MIGRATE=false` turns it off.

Two things you may want to decide differently, and I am happy to change either:

- It defaults to **on**. Opt-in would be the conservative choice for existing self-hosters, since this changes what a container does on boot.
- With several instances starting at once each one runs the check. Ecto's migration lock serialises them rather than letting them collide, but a deployment that scales out may prefer a dedicated migration step.

### Testing

Run locally against a dev instance with a real map:

- exported a map over the API, added a system and a connection to the JSON, imported it back, and confirmed in the database that the system landed at the exported position with its tag, status and labels, the connection kept its ship size type, and the pre-existing system was untouched
- connection flags covered by an integration test that drives a started map server: dangerous persists and lands in map state, bubbled is set and cleared, a new connection carries neither
- route avoidance checked against seeded rows for each combination of the two new options: dangerous only drops the flagged bridge, bubbled only drops the bubbled connection, both drop both, neither drops nothing
- pasting a system into a map that does not have it keeps its attributes - the test fails without the fix, with `status` coming back as 0
- start up schema check exercised against a database deliberately left one migration behind: it reported the pending migration, applied it, and the columns appeared; a second run reported the schema up to date and did nothing; with the switch off it skipped entirely

One thing worth flagging: the UI event router matches event names against a per handler list, and events missing from it fall through to a catch-all that logs at `debug` while the level is `info`. A new connection event therefore fails completely silently - no write, no error, nothing in the log. Cost me a while to find; a prefix match or an `info` level log there would save the next person the same hunt.

### Note

The status values in the API docs (`5: dangerousPrimary, 6: dangerousSecondary, 7: lookingFor`) do not match the enum the mapper writes (`5: target, 6: dangerous`). Nothing here depends on that any more, but the docs still look stale.


SzesciennaGeometria:
hello. iv been looking into bringing back "bubled connection" marker like pathfinder had. got stuck when potgres threw me an error since there is no such thing as "bubled" in mapper database at the moment. can you post screenshot of your "Dangerous jumpgates" marker ?

z4nr34l:
@SzesciennaGeometria pretty good catch, I've named that wrongly as we are using this mostly for ANSI that we want to avoid. But anyways it should just work fine - we're maring both sides as "dangerous" which is marking the connection RED without any special marker - and the filter is avoiding that connection in routing. Anyways, You just gave me good idea - I will look into that today and add bubbled connection markers in specific system.

P.S. Pozdrowienia od SHCK :)

SzesciennaGeometria:
coloring whole connection is an option but may be confused wyth mass/eol i think.
i was going into direction of using existing framework just to add button and an icon 
(i have button and icon ready but it does not work since this goes directly into postrgess and i havent had time to look into what can be done or is it even possible without messing up the whole DB - RFC needed) 
here is the rough visualization: 
b8ubdled2

z4nr34l:
We have currently this done as below

image
image
image

and I'm pretty sure you can tell which is which without any effort just by looking at it + added marking signatures as bubbled

image

I'm now adding indicator that will show if there is any bubbled signature in system.

SzesciennaGeometria:
those circles look sick. just thro in viariable-name in there i can later configure i my theme :P

DmitryPopov:
I'd separate import/export to a separate PR if possible

z4nr34l:
> those circles look sick. just thro in viariable-name in there i can later configure i my theme :P

I would add bubbles color, border in settings

z4nr34l:
Both bits of feedback are in.

@DmitryPopov — export/import is out of here and stands alone in #656. This branch was rebased to drop that commit; nothing else moved.

@SzesciennaGeometria — the rings are all variables now, so you can theme them:

| variable | what it sets | default |
| --- | --- | --- |
| `--rf-edge-bubble` | ring colour | `#ffb03a` |
| `--rf-edge-bubble-fill` | fill inside the ring | same colour at 18% |
| `--rf-edge-bubble-size` | radius | `24px` |
| `--rf-edge-bubble-border` | ring thickness | `2px` |
| `--rf-edge-bubble-dash` | dash pattern | `6 4` |

The radius goes through CSS too — `r` is a geometry property, so a theme sets it like any other value and the `r` attribute on the circle is only the fallback.

There is also a plain UI for it in the Connections tab of the map user settings — colour, size, border, fill — for people who would rather not keep a theme. Those live with the rest of the map user settings server-side, so they follow the user across browsers and ride along into the defaults an admin saves. Leaving a field empty emits no variable at all, so the theme keeps control; it only overrides what was actually filled in.

priv/repo/migrations/20260731111939_add_connection_dangerous_bubbled.exs
```diff
@@ -0,0 +1,25 @@
+defmodule WandererApp.Repo.Migrations.AddConnectionDangerousBubbled do
+  @moduledoc """
+  Adds the manually set connection flags: whether the connection is dangerous and which of its
+  ends are bubbled.
+
+  The generated version also re-added locked_at and locked_by_id, which are already on the table -
+  the resource snapshots are behind the database for those columns.
```
guarzo, line 7: This is right about the columns but not about the constraint, and the regenerated snapshot ends up recording something that isn't true.

`locked_at` and `locked_by_id` are indeed already on the table — `20260425000000` added them. But that migration added `locked_by_id` as a bare `:binary_id` with no `references(...)`, so the **foreign key was never created**. The old snapshot (`20251108142542.json`) mentions neither column, which is the drift you spotted.

The new snapshot fills that gap, and for `locked_by_id` it writes a full references block naming `map_chain_v1_locked_by_id_fkey` → `character_v1.id`. Since this migration doesn't create that constraint, the snapshot now claims an FK that exists in no database. Same drift as before, just pointing the other way — and a bit worse, because the next `mix ash_postgres.generate_migrations` compares the resource against the snapshot, sees the FK recorded, and emits nothing. The constraint would silently never get made.

#649 is open and does exactly this fix. It regenerates the same file, so whichever of the two lands second conflicts here either way — but if this one lands first, #649's regeneration also stops producing the migration it needs.

Cleanest path is probably to let #649 land and rebase on it, so the snapshot picks the FK up from a migration that actually creates it. Adding the `references(...)` to your own `up/0` would work too, but then you're doing #649's job in a PR about connection flags.

guarzo (commented):
Not a full review — just one thing I hit while reading the migration that I think matters before this goes further, since it overlaps with #649.

SzesciennaGeometria:
> @SzesciennaGeometria — the rings are all variables now, so you can theme them:
> variable 	what it sets 	default
> `--rf-edge-bubble` 	ring colour 	`#ffb03a`
> `--rf-edge-bubble-fill` 	fill inside the ring 	same colour at 18%
> `--rf-edge-bubble-size` 	radius 	`24px`
> `--rf-edge-bubble-border` 	ring thickness 	`2px`
> `--rf-edge-bubble-dash` 	dash pattern 	`6 4`
> 
> The radius goes through CSS too — `r` is a geometry property, so a theme sets it like any other value and the `r` attribute on the circle is only the fallback.

 sweet, cant w8 to tinker with those

z4nr34l:
Good catch, and you are right on both halves — the columns were the real drift, the constraint was not.

Fixed in `d3837fd8` by recording `locked_by_id` **without** the references block, which is what the database actually looks like after `20260425000000` added it as a bare `binary_id`. The two columns stay in the snapshot, since those genuinely exist and their absence was the drift worth closing; the key that no migration has ever created does not.

I checked the failure mode you described rather than assuming it, because it is the part that would have gone unnoticed. With the corrected snapshot, `mix ash_postgres.generate_migrations` still emits:

```elixir
alter table(:map_chain_v1) do
  modify :locked_by_id,
         references(:character_v1,
           name: "map_chain_v1_locked_by_id_fkey",
           ...
```

So #649's migration is still produced. With the references block in, that output disappears — which is exactly the silent-never-created case you were pointing at.

On ordering: this snapshot is dated `20260731111939` and #649's is `20260808152615`, so whichever lands second keeps the later snapshot as the baseline and neither swallows the other. Happy to rebase onto #649 once it lands if you would rather this file did not touch `locked_by_id` at all — but as it stands I do not think it blocks you either way.

z4nr34l:
Rebased onto `main` (`6cd0e4ba`) — conflicts resolved.

The only clash was in `map_signatures_event_handler.ex`, where #659's lazy-delete colouring fix and the bubbled-signature marker both added code to the same spot in `update_signatures`. They are independent, so both stay: #659's `updated_removed_sig_eve_ids` bookkeeping still feeds the `assign` below it, and the system-card broadcast runs after it.

Checked rather than assumed:

- Compiles, and the branch's own integration tests pass (`map_paste_systems_test`, `map_connection_flags_test`, 5 tests).
- `test/unit` gives 626 tests / 7 failures on this branch — and the same 626 / 7 on `main` at the same commit. The failures are in `SlugRecoveryTest` and `AuthControllerTest`, neither of which this branch touches, so they came with `main` rather than the rebase.

DmitryPopov:
@DanSylvest I guess you should take a look this one UI changes first.

PR #635, opened 2026-08-01
fix(api): stop a map API key from reading other maps

guarzo:
A map API key could read and change data belonging to other maps on the same server. Every `/api/v1` resource is now limited to the map the key belongs to.

Breaking: `/api/v1` no longer accepts cookie/session auth — a Bearer map API key is required, or the request returns 401.
Breaking: creating a record with a foreign `map_id` now returns 403 instead of being silently rewritten to the token's map.

@DmitryPopov @DanSylvest


guarzo:
## Proposal

## Summary

Deny-by-default Ash policies scope every route-exposed `/api/v1` resource to the token's map, closing cross-map read/write on previously-unscoped v1 resources. `/api/v1` is also now token-only — the session-auth branch wrapped the logged-in user as `ActorWithMap{map: nil}`, an actor that matched no map scope and so bypassed scoping entirely.

## How to review this (it is smaller than it looks)

The diff is ~2.5k lines, but **1.9k of that is tests** and the production diff is far more repetitive than the file count suggests:

| | lines |
|---|---|
| **Novel logic** — `policies/map_scoped.ex` + `policies/acl_scoped.ex` | **~250** |
| Same ~15-line `policies do` block applied to 18 resources | ~420 |
| Plug / controller / mix.exs | ~90 |
| Tests (104 tests) | 1,913 |

**Suggested review order — the first two items are ~90% of the actual thinking:**

1. **`lib/wanderer_app/api/policies/map_scoped.ex`** — all the real logic lives here.
2. **`lib/wanderer_app/api/map_system.ex`** — the canonical `policies do` template. Every one of the other 17 resources is this same block with a different path argument; once you've read this one, the rest is pattern-matching for deviations.
3. **`controllers/plugs/check_json_api_auth.ex`** — breaking change #1 (session auth removed).
4. **`controllers/api/map_systems_connections_controller.ex`** — hand-written, so it does *not* inherit AshJsonApi route policies and carries its own guards.
5. Skim the remaining 16 resources for anything that departs from the template.

All 18 bypass blocks are byte-identical by design, so a deviation should be visually obvious.

## Breaking changes

Both are confined to the `:api_v1` pipeline (`/api/v1/*`); no other surface is affected.

**1. Session authentication is no longer accepted on `/api/v1`.**
The plug previously fell back to the Phoenix session when no `Authorization` header was present, and wrapped the logged-in user as `ActorWithMap{map: nil}` — an actor that matched no map scope and so bypassed scoping entirely. Any client relying on cookie auth against `/api/v1` now receives **401**. A Bearer map API key is required. Session-authenticated web clients should continue to use the LiveView/internal surfaces, which are unchanged.

**2. A client-supplied foreign `map_id` on create is now rejected instead of silently corrected.**

| Request | Before | After |
|---|---|---|
| `POST /api/v1/map_systems` with no `map_id` | 201 (injected from token) | 201 — unchanged, and this is the path real clients use |
| `POST` with `map_id` equal to the token's map | 201 | 201 — unchanged |
| `POST` with a **foreign** `map_id` | 201, silently overridden by `InjectMapFromActor` | **403** via the `create_map_matches_token` policy |

Only the third row changes. Silently rewriting a caller's explicit foreign `map_id` into a success masked what was really an authorization error, so this surfaces it.

**Status-code convention.** Read/update/destroy are Ash *filter* checks, so an out-of-scope row is excluded rather than rejected: list omits it, and `GET`/`PATCH`/`DELETE` on a foreign id return **404**, not 403 — this avoids confirming that another map's row exists. Create has no existing row to filter and is guarded by a simple check, so a foreign parent returns **403**.

## Notes for reviewers

- New runtime dependency `simple_sat`: it is an *optional* dep of Ash/Crux, which resolve policy scenarios with a SAT solver. Without it any authorized action raises `UndefinedFunctionError` on `SimpleSat.solve/1`. Rationale is inline in `mix.exs`.
- `Policies.MapScoped.Trusted` uses `is_struct/2` rather than struct pattern matching on purpose — a struct pattern expands at compile time and closes a compile cycle with `Api.User` (which itself carries a policy referencing the module), failing `mix compile --force`. Rationale is inline.
- Update/destroy use **filter** checks rather than simple checks throughout. A SimpleCheck declaring `requires_original_data?` disqualifies Ash's `:atomic` bulk strategy, which breaks atomic updates (`NoMatchingBulkStrategy`) and forces the `:stream` destroy path, whose `Ash.Query.select([])` re-read returns pkey-only records that then fail Jason encoding. Explained inline in `map_connection.ex`.
- `MapSystemsConnectionsController` is hand-written and does not inherit AshJsonApi route policies, so it carries two explicit guards (path `map_id` vs token, plus actor propagation — the domain gate is `authorize :when_requested`, so a bare `Ash.read!/1` runs unauthorized). The path/token comparison normalizes both sides with `to_string/1`; a regression test asserts an upper-cased own-map id still returns 404 (normalization must not loosen into case-folding).
- The inline comments in `map_scoped.ex` are load-bearing — each documents a non-obvious Ash behavior that cost real debugging (dynamic `ref/2` not resolving through filter hydration, the atomic-strategy interaction, the compile cycle). They are why that module is 216 lines rather than 120.

## Test plan

- [x] 119/119 tests pass across `test/integration/api/v1/`, `test/wanderer_app/api/`, `test/wanderer_app_web/controllers/plugs/`
- [x] Every denial is paired with an own-map positive control, so a policy that forbade everything would not pass
- [x] `internal_authz_regression_test.exs` asserts the internal (LiveView/session) surfaces still work unauthorized-by-token — the bypass is exercised, not assumed
- [x] `authz_includes_leak_test.exs` covers scoping through JSON:API `include=` compound documents, a common way scoping gets bypassed
- [x] Full integration + web suite shows no new failures vs baseline (pre-existing failures limited to the known-flaky `auth_controller`/`slug_recovery` suites)
- [x] `mix format --check-formatted` clean; `mix compile --force` adds no new warnings

## Dependencies

None. The map-duplication fixes live in #639 and are independent of this PR — either can land first.


## Review pass — what "route-exposed" actually means here

The summary above says policies scope "every route-exposed" resource, which undersells the criterion and invites the obvious reviewer question ("why is `user` policed? it has no endpoint"). The real criterion is **`json_api do` — declaring a type**, not declaring routes.

Declaring a type is what makes a resource reachable through the API surface, as an `include=` or relationship target, whether or not it has routes of its own. Counted against the tree:

| | count | notes |
|---|---|---|
| Resources declaring `json_api do` | 18 | |
| …of those, declaring `policies do` | **18** | no gap |
| …of those, with **zero** route entries | 6 | `map`, `map_default_settings`, `map_solar_system`, `map_state`, `ship_type_info`, `user` |

Those six are policed precisely because a type with no routes is still reachable as an include target. Every remaining resource in `lib/wanderer_app/api/` — `character`, `license`, and the `*_transaction` / `*_invite` / `*_ping` / `*_webhook_subscription` / `*_jumps` set — declares no `json_api` block at all, so it is unreachable from JSON:API; a policy there would gate only internal callers. The rationale is now recorded inline in the four resources where the question is most likely to be asked.

Reproduce with:

```bash
cd lib/wanderer_app/api
comm -23 <(grep -l "json_api do" *.ex | sort) <(grep -l "policies do" *.ex | sort)   # empty
```

## Review pass — one further change

`AccessList.map_access_lists` is now `public? false`. The relationship is introduced *by this PR* purely so the `AclScoped` read policy can express `exists(map_access_lists, map_id == ^map_id)`; leaving it public would have added a policy-internal join to the derived filter surface of `/api/v1/access_lists` (the resource sets `derive_filter?(true)`). Since it has never shipped public, this is not a compatibility change — it is closing an unintended widening before it lands.

Also dropped the stale "Session Authentication" paragraph from `priv/posts/2025/07-15-api-modernization.md`, which documented the behavior breaking change #1 removes.

PR #638, opened 2026-08-01
chore(ci): move every GitHub Action off the Node 20 runtime

guarzo:
GitHub is retiring the Node 20 action runtime and our workflows still run on it. Every action in `.github/workflows/` is moved to a version that runs on Node 24, plus a one-line permissions fix so the test-results comment added in #636 can actually post.

@DmitryPopov @DanSylvest

### DIFF
--- a/.github/workflows/advanced-test.yml
+++ b/.github/workflows/advanced-test.yml
@@ -24,7 +24,7 @@ jobs:
     if: ${{ github.base_ref == 'develop' || (github.ref == 'refs/heads/develop' && github.event_name == 'push') }}
     steps:
       - name: ⬇️ Checkout repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v5
       - uses: superfly/flyctl-actions/setup-flyctl@master
 
       - name: 👀 Read app name
@@ -71,12 +71,12 @@ jobs:
           elixir-version: ${{matrix.elixir}}
         # nix build would also work here because `todos` is the default package
       - name: ⬇️ Checkout repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v5
         with:
           fetch-depth: 0
       - name: 😅 Cache deps
         id: cache-deps
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         env:
           cache-name: cache-elixir-deps
         with:
@@ -87,7 +87,7 @@ jobs:
             ${{ runner.os }}-mix-${{ matrix.elixir }}-${{ matrix.otp }}-
       - name: 😅 Cache compiled build
         id: cache-build
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         env:
           cache-name: cache-compiled-build
         with:
--- a/.github/workflows/advanced-test.yml
+++ b/.github/workflows/advanced-test.yml
@@ -27,13 +27,6 @@ jobs:
         uses: actions/checkout@v5
       - uses: superfly/flyctl-actions/setup-flyctl@master
 
-      - name: 👀 Read app name
-        uses: SebRollen/toml-action@v1.0.0
-        id: app_name
-        with:
-          file: "fly.toml"
-          field: "app"
-
       - name: 🚀 Deploy Test
         run: flyctl deploy --remote-only --wait-timeout=300 --ha=false
         env:
--- a/.github/workflows/advanced-test.yml
+++ b/.github/workflows/advanced-test.yml
@@ -24,7 +24,7 @@ jobs:
     if: ${{ github.base_ref == 'develop' || (github.ref == 'refs/heads/develop' && github.event_name == 'push') }}
     steps:
       - name: ⬇️ Checkout repo
-        uses: actions/checkout@v5
+        uses: actions/checkout@v7
       - uses: superfly/flyctl-actions/setup-flyctl@master
 
       - name: 🚀 Deploy Test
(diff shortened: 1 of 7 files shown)


DmitryPopov:
@guarzo pls resolve conflicts

guarzo:
## Proposal

Moves the whole workflow directory off the deprecated Node 20 action runtime.

## What changed

**Version bumps** (first commit) — every remaining action pin moved to its current latest major, not only the ones that were on a Node 20 runtime:

| Action | Before | After | Latest as of this PR |
|---|---|---|---|
| `actions/checkout` | v3 / v4 | **v7** | v7.0.1 |
| `actions/cache` | v3 / v4 | **v6** | v6.1.0 |
| `actions/upload-artifact` | v4 | **v7** | v7.0.1 |
| `actions/download-artifact` | v4 | **v8** | v8.0.1 |
| `actions/github-script` | v7 | **v9** | v9.0.0 |
| `docker/setup-buildx-action` | v3 | **v4** | v4.2.0 |
| `docker/setup-qemu-action` | v3 | **v4** | v4.2.0 |
| `docker/login-action` | v3 | **v4** | v4.6.0 |
| `docker/metadata-action` | v5 | **v6** | v6.2.0 |
| `docker/build-push-action` | v6 | **v7** | v7.3.0 |
| `softprops/action-gh-release` | v1 | **v3** | v3.0.2 |
| `peter-evans/find-comment` | v3 | **v4** | v4.0.0 |
| `peter-evans/create-or-update-comment` | v4 | **v5** | v5.0.0 |
| `dawidd6/action-download-artifact` | v3 | **v21** | v21 |
| `WyriHaximus/github-action-get-previous-tag` | v1 | **v2** | v2.0.0 |
| `appleboy/ssh-action` | v1.0.3 | **v1.2.5** | v1.2.5 |

`erlef/setup-beam@v1` is unchanged — v1 is still the current major.

Two notes on style. Everything except `appleboy/ssh-action` is pinned to a floating major tag, matching what the directory already did; `ssh-action` was already pinned to an exact patch and stays that way, since it is the one action here that runs against the deploy host. And the "Latest" column is the value `gh api repos/<owner>/<repo>/releases/latest` returned while preparing this — worth re-checking if the PR sits for a while.

**Third-party removals** (remaining commits) — four actions with no maintained Node 24 release, replaced with inline shell:

- `SebRollen/toml-action` — the step's output was never consumed; deleted rather than replaced.
- `markpatterson27/markdown-to-output` + `2428392/gh-truncate-string-action` — the changelog chain is now a single inlined step.
- `tsickert/discord-webhook` — inlined as `jq | curl`.

The inlined `jq | curl` steps set `pipefail` so a failed webhook post fails the step rather than being masked by `curl`'s exit code.

## Verification

- `actionlint` on this branch vs `main`: **zero new findings, 9 resolved** (all of them "runner ... is too old to run on GitHub Actions"). The ~62 remaining are pre-existing `SC2086` shellcheck notes and one undefined-property warning in `advanced-test.yml`, untouched here.
- All six workflow files parse as valid YAML.
- No action reference anywhere in `.github/workflows/` is still on a Node 20 runtime.

The build and release paths can't be exercised from a PR, so those are verified by inspection and lint only.


## Review pass

The first version of this PR bumped only the actions that were on the deprecated Node 20 runtime, which left `github-script`, both `peter-evans/*` actions, `dawidd6/action-download-artifact`, and `appleboy/ssh-action` behind on older majors. They are all current now — the table above is the full set, verified against each repository's `releases/latest` rather than from memory.

`actionlint` was re-run after the additional bumps, comparing every workflow file at `main` against this branch. The result is unchanged: the six shellcheck classes are byte-for-byte identical on both sides (`SC2034`×1, `SC2046`×4, `SC2086`×48, `SC2129`×3, `SC2216`×1, `SC2320`×1, plus 4 expression warnings), and the 9 "too old" findings present on `main` are all gone. **Zero new findings, 9 resolved.**




## Also here: the test-results comment permissions (`pr-comment.yml`)

Rebasing on `main` pulled `pr-comment.yml` (from #636) into this PR's diff for the `find-comment` / `create-or-update-comment` bumps above, which surfaced a second problem in the same file. Fixing it here rather than in a separate PR since the file is already touched — happy to split it out if you'd rather.

#636 did the hard part correctly: `workflow_run` gets a writable token for fork PRs, and the job log confirms `Issues: write` is granted. But the POST still 403s, and has on every run since #636 merged:

```
POST /repos/wanderer-industries/wanderer/issues/638/comments
403 Resource not accessible by integration
x-accepted-github-permissions: issues=write; pull_requests=write
```

Commenting on a *pull request* needs the pull-requests scope too, even though the body goes to the issue-comments endpoint. #636 set `pull-requests: read` on the reasoning that the comment is an issue comment — true for issues, not for PRs. The pre-#636 `test.yml` had `pull-requests: write`.

Fix is one line: `pull-requests: read` → `write`.

Two caveats:

- **Not verifiable before merge.** `workflow_run` always runs the default branch's copy of the file, so the run triggered by this PR will still fail with the same 403 using main's `read`. First real test is the next PR opened after this lands.
- This has never worked, so there is no regression to compare against — every PR in the repo comes from a fork, so the comment was already failing (silently, read-only token) before #636 changed the mechanism.

Separately, and not addressed here: the rendered body from this PR's run reports 13/927 tests failed, `mix format` failing, and 258 Credo issues, while the Test Suite job still reports pass — those steps are all `continue-on-error: true`, so the "Quality Gates" section is informational and nothing enforces the stated limits. #642 looks like the right place for that.

PR #639, opened 2026-08-01
fix(map): duplicating a map fails with an error

guarzo:
Duplicating a map fails with a server error, and before that it silently produced copies with no signatures in them. Both are fixed.

Merge this before #642.

@DmitryPopov @DanSylvest

### DIFF
--- a/lib/wanderer_app/map/operations/duplication.ex
+++ b/lib/wanderer_app/map/operations/duplication.ex
@@ -94,28 +94,11 @@ defmodule WandererApp.Map.Operations.Duplication do
 
   # Copy a single system
   defp copy_single_system(source_system, new_map_id) do
-    # Get all attributes from the source system, excluding system-managed fields and metadata
-    excluded_fields = [
-      # System managed fields
-      :id,
-      :inserted_at,
-      :updated_at,
-      :map_id,
-      :map,
-      # Ash/Ecto metadata fields
-      :__meta__,
-      :__lateral_join_source__,
-      :__metadata__,
-      :__order__,
-      :aggregates,
-      :calculations
-    ]
-
-    # Convert the source system struct to a map and filter out excluded fields
+    # Same allowlist approach as connections -- see acceptable_attrs/3 below for
+    # why a denylist was the wrong shape here.
     system_attrs =
       source_system
-      |> Map.from_struct()
-      |> Map.drop(excluded_fields)
+      |> acceptable_attrs(MapSystem, :create)
       |> Map.put(:map_id, new_map_id)
 
     MapSystem.create(system_attrs)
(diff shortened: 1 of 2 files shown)


guarzo:
## Proposal

## Summary

Two independent bugs in `WandererApp.Map.Operations.Duplication`. Both reproduce on `main` today — the existing duplication suite has **6 failing tests** at `main`, and this branch takes it to 39/39.

## 1. Every map duplication returned 500

`copy_single_system/2` and `copy_single_connection/3` built their create params with a **denylist**:

```elixir
source |> Map.from_struct() |> Map.drop(excluded_fields)
```

Every attribute not explicitly named was forwarded to `create`. When 2f86fc9f added `locked_at` / `locked_by` / `locked_by_id` to `MapConnection` without adding them to the `:create` accept list, duplication started failing with `Ash.Error.Invalid.NoSuchInput`.

Replaced with `acceptable_attrs/3`, which derives an **allowlist from the target action's own `:accept` list**. A newly added attribute is copied only if the action accepts it, so the two cannot drift apart again — this class of regression can't recur.

## 2. Signatures were never copied

`get_all_map_signatures/2` called `MapSystemSignature.by_system_id_all/1` with `%{system_id: id}` instead of a bare id, and mapped the resulting error to `[]`:

```elixir
case MapSystemSignature.by_system_id_all(%{system_id: system_id}) do
  {:ok, signatures} -> signatures
  {:error, _} -> []          # <- swallowed the arity/shape error
end
```

So every duplication with `copy_signatures: true` produced a map with **zero signatures while reporting success**. Read failures now propagate as `{:error, {:signature_read_failed, reason}}` rather than being dropped — a partial copy can no longer masquerade as a complete one.

The read now goes through `by_system_id/1` rather than `by_system_id_all/1`. The `_all` variant includes soft-deleted rows, so fixing only the argument shape would have copied signatures the user had already deleted into the new map. A regression test covers that case.

## 3. Dead remapping code that would have corrupted connections if it ever ran

`copy_single_connection/3` piped its attributes through `update_system_references/2` → `maybe_update_system_reference/3`, which looked each endpoint up in `system_mapping`:

```elixir
case Map.get(system_mapping, old_system_id) do
  nil -> attrs                              # keep original
  new_system_id -> Map.put(attrs, field, new_system_id)
end
```

`system_mapping` is keyed by **`MapSystem` UUIDs**, while `solar_system_source` / `solar_system_target` are declared `:integer` on `MapConnection` — EVE solar-system ids. An integer key can never match a UUID key, so the lookup always returned `nil` and the function was a guaranteed no-op.

It is removed rather than repaired, because there is nothing to repair: EVE solar-system ids are identical between the source map and its duplicate, so connection endpoints must not be remapped at all. Had the mapping ever matched, it would have overwritten valid solar-system ids with `MapSystem` UUIDs and corrupted every copied connection. The remaining use of `system_mapping` — translating `MapSystem` UUIDs for signature references — is genuine and untouched. A comment on `copy_single_connection/3` now records why the two identifier spaces are different.

## Test plan

- [x] `test/integration/map_duplication_api_controller_success_test.exs`, `test/unit/map_duplication_test.exs`, `test/unit/map_duplication_api_test.exs`: **39/39 pass** (6 failures on `main` before this change)
- [x] Two new tests cover signature copying, which no test asserted before — the existing `copy_signatures => true` test only checked for a 201 and a new id, so it passed against the broken code
- [x] Both new tests verified to **fail** against the previous implementation (reverted the lib change and confirmed red), so they are non-vacuous
- [x] `mix format --check-formatted` clean; `mix compile` adds no new warnings

## Dependencies

None to land. Note that **#642 depends on this** — the 6 duplication failures remaining on that branch are exactly the `NoSuchInput` bug fixed here, so this should merge first.

PR #640, opened 2026-08-01
fix(ci): make the test job capable of failing, and fix the 16 failures it was hiding

guarzo:
## Summary

CI's test job has been **structurally incapable of reporting failure**. Behind that mask, 22 tests were failing on `main`. This PR fixes 16 of them and then turns the gate on.

Please read the CI change with that framing: **this will turn the build red if anything regresses.** That is the point. Nothing here makes the suite stricter than it should be — it makes a green check mean something.

## Problem 1 — CI could not fail

Three independent defects in `.github/workflows/test.yml`:

1. `output=$(mix test --cover 2>&1 || true)` discarded the exit status. The `exit_code=$?` two lines below measured the preceding `echo`, so it was always `0`.
2. `continue-on-error: true` on the test step meant that even when the workflow's own parser correctly computed `status=❌ Some Failed`, the job still reported green. The failure was detected and then deliberately discarded.
3. Integration tests never ran **anywhere** in CI. `test/test_helper.exs` excludes `:integration` by default and no workflow passed `--include`. ~64 tests have never executed.

### What changed

- The test step captures the real exit status and exits with it.
- `continue-on-error` removed **from the test step only**. It is deliberately kept on the advisory reporting steps (coverage, Credo, Dialyzer, summary) — those should not gate a merge. Those steps and the PR-comment steps gained `if: always()` so the results comment is still posted when the suite is red, which is when it's most useful.
- Integration tests run in a **separate `integration-test` job** rather than by changing the default exclusion, so local `mix test` stays fast.

### Verified, not reasoned about

With a deliberately failing test added locally and the step's shell replicated verbatim:

```
new logic: status=FAILED, test_exit_code=2   -> job fails
old logic: exit_code=0                       -> job green
```

The temporary test was then removed.

## Problem 2 — the failures CI was hiding

Measured against `b7ddbc48`, four seeds (`0`, `42`, `1`, `12345`), comparing **sets, not counts**:

| | baseline | this branch |
|---|---|---|
| union across 4 seeds | **22** | **6** |
| new failures introduced | — | **0** |

The 6 remaining are all `MapDuplicationAPIControllerSuccessTest`, caused by `Map.Operations.Duplication` passing `locked_at`/`locked_by_id`/`locked_by` to `MapConnection.create`, which does not accept them. **That is already fixed in #635** (`acceptable_attrs/3` allowlist), so it is deliberately not duplicated here.

> ⚠️ **Merge order matters:** please merge **#635 first**. Merging this PR alone turns CI red on those 6 tests.

### Root causes found

Each was instrumented, not guessed — the error message pointed away from the real cause in two of three.

**1. `maps_v1.scopes` had a corrupt column default (production bug).**
Two migrations declared `default: '{wormholes}'`. A single-quoted Elixir literal is a **charlist**, so the generated DDL stored an eleven-element text array of the *character codes* of the string `{wormholes}` rather than the one-element array `{wormholes}`. Any row inserted without an explicit `scopes` got garbage, and every later read failed with `Ash.Error.Unknown` (`cannot load ... as type {:array, Ash.Type.Atom}`).

This surfaced as a *slug recovery* failure, which is why it went undiagnosed — the failing assertion was about duplicate slugs, but the actual error came from re-reading a fixture row. Fixed with a corrective migration (sets the proper default and repairs affected rows) plus a fix to `migration_defaults` so regeneration doesn't reintroduce it.

**2. Fake stargates leaked into the global jump cache (10 order-dependent failures).**
`map_scope_filtering_test.exs` seeds two fabricated `jump_*` entries into the global cache and never removed them. `CachedInfo.get_solar_system_jump/2` only rebuilds the index when a key is *missing*, so they persist for the rest of the run — and they cover the exact system IDs `MapScopesTest` uses for its "no stargate exists" cases.

Proven by instrumenting the assertion: in isolation the lookup returns `{:ok, nil}`, in the full run it returns a jump struct — while the database contains **no such row** (verified by direct count). Running the same file set with `--max-cases 1` was clean, confirming a shared-cache race rather than a plain ordering bug.

**3. `MapScopesTest` leaked stub records into `:system_static_info_cache`.**
Its stubs carry only `solar_system_id`/`system_class` and overwrote system `30_000_142` (Jita), which `CommonAPIControllerTest` and `OpenAPIValidationTest` populate fully. Whichever ran second lost. Bisected to a minimal reproducing pair:

```
mix test .../common_api_controller_test.exs .../map_scopes_test.exs --seed 42   # 1 failure
mix test .../common_api_controller_test.exs --seed 42                           # clean
mix test .../map_scopes_test.exs --seed 42                                      # clean
```

No assertion was weakened or skipped to make anything pass.

## Commits

Production and test fixes are kept separate:

- `fix(db)` — repair `maps_v1.scopes` charlist default *(production)*
- `test(auth)` — give AuthController tests a session and flash *(test-only; the controller legitimately expects the browser pipeline)*
- `test(map)` — stop `MapScopesTest` leaking system static info cache
- `test(map)` — stop leaking fake stargates into the global jump cache
- `ci` — let the test job actually fail, and run integration tests

## Question for maintainers

**Should the ~64 integration tests gate merges?** They have never run in CI, so making them a hard gate is a real policy change. I've marked the job `continue-on-error` so you get visibility first — remove that line to promote it to a gate. Happy to do either.


guarzo:
Follow-up commit `fdf78ee2` after a self-review:

- `@bad_default` is now **derived** from the offending literal (`~c"{wormholes}"`) rather than hand-transcribed as `~w(123 119 ...)`. A single transcription typo would have made the `WHERE` clause match nothing and silently repair zero rows — the failure mode would have been invisible. Verified the emitted SQL is byte-identical to before.
- Documented why `down/0` is deliberately **not** a mirror of `up/0`: restoring the original charlist default would reintroduce the bug, so it drops the default instead, and already-repaired rows stay repaired. Rolling back leaves the schema in a different-but-correct state.
- Confirmed by grep that both `20260331192521` and `20260406213852` used exactly the same literal, so the single value catches rows written by either.

Re-verified after the change: migration rolls back and re-applies cleanly, and the full suite is still **6 failures at all four seeds** (0/42/1/12345), all `MapDuplicationAPIControllerSuccessTest` — i.e. only the set owned by #635.

One known duplication I deliberately did **not** fix: the new `integration-test` job repeats the checkout/setup/cache/deps/db-setup sequence from the `test` job. The repo has no `.github/actions/` and all five existing workflows already duplicate this setup, so extracting a composite action would break with convention and widen this PR's scope. Happy to do it separately if you'd prefer.

PR #642, opened 2026-08-02
fix(ci): make the test job fail when tests fail

guarzo:
The test job always reported green even when tests failed, and behind that 22 tests were broken on `main`. This makes the job report failure and fixes 16 of them, including a corrupt database default on `maps_v1.scopes` that affected production.

Merge #639 first — the 6 failures left after this PR are the duplication bug it fixes.
Open question for @dedo1911: should the ~64 integration tests gate merges? They are non-blocking here.

@DmitryPopov @DanSylvest


dedo1911:
Thanks for digging into this — the CI analysis is right, and the charlist default is a genuine production bug. A few things before this can go in.

**Merge order: the prerequisite is #639, not #635.** #635 doesn't touch the duplication path at all, and `acceptable_attrs` doesn't exist in it (or anywhere in the tree yet). The fix for those 6 `MapDuplicationAPIControllerSuccessTest` failures is #639, which changes `duplication.ex` and adds the allowlist. Root cause confirmed on main: the denylist in `copy_single_connection/3` doesn't exclude `locked_at` / `locked_by_id` / `locked_by`, and `MapConnection.create` doesn't accept them.

**On your question — no, let's not gate the integration tests yet.** That job is already red for reasons unrelated to this PR: 8 failures in `CharacterLocationTrackingTest` (`Timeout waiting for map ... to start`), which is exactly what #637 fixes. So keep `continue-on-error` here, let #637 land, then drop the line in a follow-up. I'm happy to gate them once they're green — I just don't want to turn them into a hard gate while they're failing.

**Two things to change in this PR:**

1. The Ash resource snapshot wasn't regenerated. `priv/resource_snapshots/repo/maps_v1/20260406213852.json` still has `"default": "'{wormholes}'"` while `map.ex` now says `["wormholes"]`, so the next `mix ash_postgres.generate_migrations` will emit a duplicate migration on top of your corrective one. Please regenerate the snapshot so the resource and the snapshot agree.

2. The integration job runs `mix test --include integration`, which re-runs the entire suite — the log shows `926 tests, 14 failures, 52 skipped`, the same total as the main test job. `--only integration` runs just the ~64 tagged ones, which avoids duplicating ~10 minutes of CI and stops the job from re-reporting the 6 duplication failures the other job already covers.

FYI the `mix format --check-formatted` failure is pre-existing on main (it fails on 544467ab too) and that step doesn't gate the job, so no need to address it here.

Plan on my side: #639 → #637 → this one, probably with a rebase since #635 also touches `lib/wanderer_app/api/map.ex`.

guarzo:
> Thanks for digging into this — the CI analysis is right, and the charlist default is a genuine production bug. A few things before this can go in.
> 
> **Merge order: the prerequisite is #639, not #635.** #635 doesn't touch the duplication path at all, and `acceptable_attrs` doesn't exist in it (or anywhere in the tree yet). The fix for those 6 `MapDuplicationAPIControllerSuccessTest` failures is #639, which changes `duplication.ex` and adds the allowlist. Root cause confirmed on main: the denylist in `copy_single_connection/3` doesn't exclude `locked_at` / `locked_by_id` / `locked_by`, and `MapConnection.create` doesn't accept them.
> 
> **On your question — no, let's not gate the integration tests yet.** That job is already red for reasons unrelated to this PR: 8 failures in `CharacterLocationTrackingTest` (`Timeout waiting for map ... to start`), which is exactly what #637 fixes. So keep `continue-on-error` here, let #637 land, then drop the line in a follow-up. I'm happy to gate them once they're green — I just don't want to turn them into a hard gate while they're failing.
> 
> **Two things to change in this PR:**
> 
> 1. The Ash resource snapshot wasn't regenerated. `priv/resource_snapshots/repo/maps_v1/20260406213852.json` still has `"default": "'{wormholes}'"` while `map.ex` now says `["wormholes"]`, so the next `mix ash_postgres.generate_migrations` will emit a duplicate migration on top of your corrective one. Please regenerate the snapshot so the resource and the snapshot agree.
> 2. The integration job runs `mix test --include integration`, which re-runs the entire suite — the log shows `926 tests, 14 failures, 52 skipped`, the same total as the main test job. `--only integration` runs just the ~64 tagged ones, which avoids duplicating ~10 minutes of CI and stops the job from re-reporting the 6 duplication failures the other job already covers.
> 
> FYI the `mix format --check-formatted` failure is pre-existing on main (it fails on [544467a](https://github.com/wanderer-industries/wanderer/commit/544467ab34cecffa5d1ce3050cd1028413cb64bb) too) and that step doesn't gate the job, so no need to address it here.
> 
> Plan on my side: #639 → #637 → this one, probably with a rebase since #635 also touches `lib/wanderer_app/api/map.ex`.

Yeah, sorry -- I ended up splitting 639 out of 635 as it didn't really make sense to combine the two things

guarzo:
Both changes pushed.

**1. Snapshot regenerated.** Ran `mix ash.codegen` and kept only the `maps_v1` snapshot, named `20260801200000.json` to match the existing corrective migration rather than the codegen timestamp — that migration already does the schema change and the data repair, so I dropped the generated migration instead of adding a second one. The snapshot delta is just the two lines:

```
-      "default": "'{wormholes}'",
+      "default": "[\"wormholes\"]",
-      "hash": "17E507A2...",
+      "hash": "F00091B5...",
```

`mix ash.codegen --dry-run` now emits nothing for `maps_v1`, so the duplicate-migration problem is gone.

**2. `--only integration`.** Changed, with a comment explaining why `--include` was wrong. There are 64 `:integration` tags in `test/`, matching your count.

**One thing codegen surfaced that I did *not* touch:** `map_chain_v1` has the same class of drift, pre-existing on `main`. `20260425000000_add_map_connection_locked_by.exs` added `locked_at` and `locked_by_id` by hand without regenerating the snapshot, so `mix ash.codegen` on a clean `main` still wants to emit an `add :locked_at` / `add :locked_by_id` migration. Out of scope here — happy to send it as a separate one-file PR if you want it.

Merge order noted: #639 → #637 → this one. I'll rebase once those land.

guarzo:
Followed up on the `map_chain_v1` drift I mentioned above — it is #649. It turned out to be more than a stale snapshot: `20260425000000` added `locked_by_id` as a bare `:binary_id`, so the FK behind `belongs_to :locked_by` was never actually created. Verified end to end against a scratch database, including the orphaned-row path.

Also checked the rest of my open PRs for the same pattern while I was at it: #646 had it (fixed there, and it would have silently dropped `ON DELETE` actions on three FKs), #635 is clean.

DmitryPopov:
@guarzo conflicts now

guarzo:
## Proposal

## Summary

CI's test job is structurally incapable of reporting failure, and behind that mask 22 tests are failing on `main`. This fixes 16 of them and then turns the gate on.

Read the CI change with that framing: **this will turn the build red if anything regresses.** That is the point.

## Problem 1 — CI could not fail

Three independent defects in `.github/workflows/test.yml`:

1. `output=$(mix test --cover 2>&1 || true)` discards the exit status. The `exit_code=$?` two lines below measures the preceding `echo`, so it is always `0`.
2. `continue-on-error: true` on the test step means that even when the workflow's own parser correctly computes `status=❌ Some Failed`, the job still reports green. The failure is detected and then discarded.
3. Integration tests never run **anywhere** in CI. `test/test_helper.exs` excludes `:integration` by default and no workflow overrides it. ~64 tests have never executed.

### What changed

- The test step captures the real exit status and exits with it.
- `continue-on-error` removed **from the test step only**. It stays on the advisory reporting steps (coverage, Credo, Dialyzer, summary) — those shouldn't gate a merge. Those steps and the PR-comment steps gained `if: always()` so the results comment is still posted when the suite is red, which is when it's most useful.
- Integration tests run in a separate `integration-test` job via `mix test --only integration`, so local `mix test` stays fast and the job doesn't re-run the ~926 tests the main job already covers.

Verified rather than reasoned about — with a deliberately failing test added locally and the step's shell replicated verbatim:

```
new logic: status=FAILED, test_exit_code=2   -> job fails
old logic: exit_code=0                       -> job green
```

## Problem 2 — the failures CI was hiding

Measured against `b7ddbc48`, four seeds (`0`, `42`, `1`, `12345`), comparing **sets, not counts**:

| | baseline | this branch |
|---|---|---|
| union across 4 seeds | **22** | **6** |
| new failures introduced | — | **0** |

### Root causes fixed here

Each was instrumented, not guessed — the error message pointed away from the real cause in two of three.

**1. `maps_v1.scopes` had a corrupt column default (production bug).**
Two migrations declared `default: '{wormholes}'`. A single-quoted Elixir literal is a **charlist**, so the generated DDL stored an eleven-element text array of the *character codes* of the string `{wormholes}` rather than the one-element array `{wormholes}`. Any row inserted without an explicit `scopes` got garbage, and every later read failed with `Ash.Error.Unknown` (`cannot load ... as type {:array, Ash.Type.Atom}`).

This surfaced as a *slug recovery* failure, which is why it went undiagnosed — the failing assertion was about duplicate slugs, but the actual error came from re-reading a fixture row. Fixed with a corrective migration (sets the proper default and repairs affected rows) plus a fix to `migration_defaults` so regeneration doesn't reintroduce it.

**2. Fake stargates leaked into the global jump cache (10 order-dependent failures).**
`map_scope_filtering_test.exs` seeds two fabricated `jump_*` entries into the global cache and never removes them. `CachedInfo.get_solar_system_jump/2` only rebuilds the index when a key is *missing*, so they persist for the rest of the run — and they cover the exact system IDs `MapScopesTest` uses for its "no stargate exists" cases.

Proven by instrumenting the assertion: in isolation the lookup returns `{:ok, nil}`, in the full run it returns a jump struct — while the database contains **no such row** (verified by direct count). The same file set with `--max-cases 1` is clean, confirming a shared-cache race rather than a plain ordering bug.

**3. `MapScopesTest` leaked stub records into `:system_static_info_cache`.**
Its stubs carry only `solar_system_id`/`system_class` and overwrote system `30_000_142` (Jita), which `CommonAPIControllerTest` and `OpenAPIValidationTest` populate fully. Whichever ran second lost. Bisected to a minimal reproducing pair:

```
mix test .../common_api_controller_test.exs .../map_scopes_test.exs --seed 42   # 1 failure
mix test .../common_api_controller_test.exs --seed 42                           # clean
mix test .../map_scopes_test.exs --seed 42                                      # clean
```

No assertion was weakened or skipped to make anything pass.

## Dependencies

**Merge #639 first.** The 6 remaining failures are all `MapDuplicationAPIControllerSuccessTest`, caused by `Map.Operations.Duplication` passing `locked_at`/`locked_by_id`/`locked_by` to `MapConnection.create`, which doesn't accept them. #639 fixes that with the `acceptable_attrs/3` allowlist, so it isn't duplicated here — but merging this PR alone turns CI red on those 6 tests.

## Open question for @dedo1911

**Should the ~64 integration tests gate merges?** They have never run in CI, so making them a hard gate is a real policy change. The job is marked `continue-on-error` so you get visibility first — remove that line to promote it to a gate. Happy to do either.


---

## Review pass

Follow-up from a self-review against the upstream pre-submit checklist, in `c0379295`:

- **The `integration-test` job is renamed `Integration Tests`** (from `Integration Tests (non-blocking)`). Encoding a temporary policy in the check *name* means that promoting the job to a gate later would rename the check, and branch protection matches checks by name — so the rename would silently stop enforcing it. The name is now stable and the temporary state lives in a comment on `continue-on-error` instead, which is the line you actually delete to promote it.

### Overlap with #650

#650 also edits `.github/workflows/test.yml`. **They do not conflict** — verified with `git merge-tree`, which produces no conflict markers in either order. #650's hunks are in the setup and codegen-check region (lines ~95–335); this PR's are the test-step exit handling and the `integration-test` job at the end of the file. Either can merge first.

dedo1911:
#637 landed this morning (f64d34ff), so my answer to the open question changes from "not yet" to **yes — after one green run.**

The evidence that it's ready: the last integration job run on this branch (`c0379295`) reported

```
926 tests, 9 failures, 862 excluded, 30 skipped
```

and all 9 of those failures were in `test/integration/map/character_location_tracking_test.exs`. That is exactly the set #637 fixes — "the 8 reported, plus an ordering-dependent 9th". Nothing else in the tagged set was red.

So: merge main in, let CI run once. If **Integration Tests** comes back green, drop `continue-on-error` and the `# Temporary, pending #637` comment **in this PR**, not a follow-up — shipping a flag we already know we're deleting is churn, and it keeps the gate decision in the PR that introduces the job. If something unexpected is red, keep the line and we promote separately.

**One thing to know before gating it:** 30 of the 64 tests `--only integration` selects are `@tag :skip` — 25 in `map_pool_crash_integration_test.exs`, 5 in `character_tracking_enable_test.exs`. So the gate actually covers 34 executing tests, not 64. Doesn't change the answer, but "~64 integration tests" is optimistic, and those 30 deserve their own look at some point.

**On the conflict** (thanks @DmitryPopov): resolved locally, it is only `.github/workflows/test.yml`. main's b73f9f31 replaced the peter-evans find-comment/create-or-update-comment pair with a "Render PR comment body" + upload-artifact pair posted by the privileged `pr-comment.yml` via `workflow_run`, and dropped the `pull-requests`/`issues` write scopes accordingly. Your side of that hunk was only `if: always()` on the two peter-evans steps, and main's replacement steps already carry `if: always()` — so main's side is taken wholesale and your intent is preserved. The `integration-test` job is appended after the rewritten reporting steps. The rest of the file merged clean: real exit status on the test step, `continue-on-error` removed from it, `if: always()` on the advisory steps.

Maintainer-can-modify is on, so I can push the merge to your branch if that saves you a round trip — otherwise it's yours to rebase.

**Merge order otherwise unchanged:** #639 is still open, so the 6 `MapDuplicationAPIControllerSuccessTest` failures are still there and it still goes first. #637 done → #639 → this one.

guarzo:
Merged main in. Only conflict was test.yml: took main's version of the comment-posting steps, re-appended the integration job after them. The exit-code fix and the `if: always()` lines came through clean.

Leaving `continue-on-error` until Integration Tests comes back green, then dropping it here rather than in a follow-up. #639 still goes first.

guarzo:
Integration Tests came back clean on the merge commit: `927 tests, 0 failures, 862 excluded, 30 skipped`. So `continue-on-error` is gone and the job is a real gate now. Check name is unchanged. Turned out to be 35 executing rather than 34, so I put that number in the comment.

Test Suite is still red at 6/927, which is the #639 dependency, unchanged.

One thing worth fixing separately: the test step captures `mix test` output into a variable and writes it to `test_output.txt`, which never gets echoed or uploaded. Now that the step actually fails the job, a red run gives you a failed check with no test output in the log. I only got the 6/927 count out of the pr-comment artifact. Happy to add the upload here or in its own PR, whichever you prefer.

Also `test/support/openapi_spec_analyzer.ex` is unformatted on main already, so the formatting row is red for reasons unrelated to this branch.

PR #644, opened 2026-08-02
feat(discord): per-map Discord kill notifications

guarzo:
## Summary

Adds per-map Discord kill notifications. Map owners and admins configure a
Discord webhook under **Map settings → Notifications**; kills detected on that
map's systems are posted to the channel as embeds, optionally restricted to
wormhole space and excluding chosen systems.

The filter is server-side and per-map — deliberately independent of the
per-user filters on the in-app kills widget.

## What's in it

**Data** — `MapDiscordNotification` Ash resource with an AshCloak-encrypted
webhook URL (only ever re-displayed as a masked hint), cascade delete with the
map, and a migration.

**Delivery** — `DiscordDispatcher` (GenServer) filters and dedups kills, then
hands each map's queue to a per-map `Discord.Worker` under a
`DynamicSupervisor`/`Registry`. The worker handles Discord's 10-embeds-per-
message cap, honours `retry_after` on 429, disables the webhook immediately on
404, and disables after 10 consecutive failures. HTTP goes through an
`HttpClient` seam over an isolated Finch pool (`WandererApp.Finch.Discord`), so
Discord traffic cannot starve the app's other pools.

**UI** — `MapNotificationsComponent` LiveView tab with URL entry, wormhole-only
toggle, excluded-systems picker, and a "Send test message". Delivery outcomes
are written by the map's worker, in a different process from whoever has the tab
open, so `record_success`/`record_failure`/`disable` broadcast on a per-map topic
after the transaction commits; the settings LiveView subscribes once per page and
refreshes the component via `send_update/2`. Without it a webhook that had been
auto-disabled after ten failures would keep rendering as healthy until the tab
was reopened.

**Config** — `WANDERER_DISCORD_POOL_SIZE` (default `10`) sizes the Finch pool.
The whole feature is gated on the existing `WANDERER_WEBHOOKS_ENABLED`; when it
is false the workers do not start and the test button reports that
notifications are disabled on this server.

**Docs** —  `.env.example` entry, and an announcement
post at `priv/posts/2026/08-02-discord-kill-notifications.md`.

## Incidental fixes

- `core_components.ex`: `live_select/1` now declares the `id` and `phx-target`
  attrs it was already being passed, and forwards them only when non-nil.
  Behaviour-preserving for the three existing call sites, which pass neither.
- `discord_dispatcher_test.exs`: removed two hidden dependencies — the tests
  relied on static system data that `mix test` never seeds (now seeded into
  `:system_static_info_cache`), and the negative assertions were vacuous because
  delivery happens in an async task the old `settle/1` could not observe (now
  `refute_delivery/2`, which waits for the worker to go idle). That cache is
  global and not sandboxed per test, and several suites seed the full Jita row
  and never clean it up, so `on_exit` restores whatever value was there before
  rather than deleting the key.

## Verification

- Feature test set: **123 tests, 0 failures**.
- Full suite: **1040 tests, 13–15 failures depending on `--seed`**. Thirteen are
  constant (`AuthControllerTest` ×6, `MapDuplicationAPIControllerSuccessTest`
  ×6, `SlugRecoveryTest` ×1); the rest come and go with the ordering
  (`CommonAPIControllerTest`, `OpenAPIValidationTest` on this branch,
  `CommonAPIControllerTest`, `MapScopesTest` on the merge-base).
- That variance is pre-existing, not introduced here. The merge-base was run in
  a detached worktree across seeds 222/555/777/999 and produced **13–23
  failures** in the same modules; the branch's 13–15 sits inside that range. The
  flaky family shares one signature — a partial `map_solar_system` Jita row
  (`%{"solar_system_id" => 30000142, "system_class" => 7}`, no
  `solar_system_name`) left in a global cache by an earlier test — which
  reproduces on `main`. Note that seeds are not comparable between the two
  trees: this branch adds 114 tests, so the same seed yields a different
  execution order.
- The test fixes were mutation-checked: flipping the behaviour under test now
  fails the test, which it did not before. The same was done for the
  status-topic unsubscribe test — with the `Phoenix.PubSub.unsubscribe/2` call
  deleted it fails, which is why it reads the subscriber list directly rather
  than relying on rendered output (the component's own `loaded_map_id` guard
  masks a leaked subscription).
- `mix format --check-formatted` is clean on every file this branch touches.
  Five unformatted files elsewhere in the repo are pre-existing and untouched.
- `--warnings-as-errors` cannot pass on this repo's baseline: 7 warnings remain
  in files this branch does not touch. The 2 warnings that did originate here
  are fixed.

## Why the Notifications tab lives in Map settings

The webhook URL is a map-level credential, so it sits in the
`/maps/:slug/settings` modal rather than the in-map **Admin Settings** tab.
That route is admin-only in its entirety: `apply_action` gates it on
`Maps.check_user_can_delete_map/2`, which returns `:not_authorized` unless the
caller's resolved permissions include `delete_map`, and the entry button is
separately gated on `Maps.can_edit?/2` (owner or `:admin` role). Every sibling
tab there — General, Balance, Subscription, Import/Export, Public API, Bots —
is already admin-only, so there is no user-facing tab for this to sit beside.

The in-map `MapSettings` widget is the opposite case: any map member opens it,
and `AdminSettings.tsx` is one gated tab among per-user display preferences.
Putting a credential there would rest on a single tab's gate rather than a
route-level one.

The tab is hidden outright when `WANDERER_WEBHOOKS_ENABLED` is false, following
the `:if={@map_subscriptions_enabled?}` pattern the Balance and Bots tabs use —
on a deployment that never enables webhooks, users are not offered a webhook
form whose test button can only report the feature is off. The guard is on the
content pane as well as the tab strip, since `change_settings_tab` assigns
whatever tab string the client sends.

## Screenshots

Unconfigured notifications tab
notifications

Notifications enabled
notificationsfull

Example notifications
killexample


guarzo:
Superseded by a four-PR stack — same feature, split so the schema and the delivery policy can be reviewed separately:

- #651 — extract the wormhole class list into `SystemClass` (standalone, no dependencies)
- #652 — `MapDiscordNotification` resource + migration + snapshot
- #653 — delivery pipeline: worker, embed formatter, HTTP client, dispatcher
- #654 — map settings tab + docs

#651 is ready to review now. The other three are drafts because GitHub can't base a cross-fork PR on a fork branch, so each shows a cumulative diff until its parent merges; each body names the files that are actually new to it.

#653 is where the design decisions live — retry and failure classification, per-event success semantics, queue bounds, supervision ordering.

Closing this in favor of the stack.

PR #645, opened 2026-08-03
Add spawn type and direction options to auto bookmarks

leesolway:
Added two sets of template vars

1. Direction - indicating if a wormhole is incoming or outgoing
2. Spawn Type - Static, Wandering or K162
3. Fixed an issue where customMapping fell back to an inline {} when settings.bookmark_custom_mapping was unset

This plugs a gap in how we name wormholes.

image
image

### DIFF
--- a/assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/components/BookmarkNameFormatSetting.tsx
+++ b/assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/components/BookmarkNameFormatSetting.tsx
@@ -40,7 +40,7 @@ const VARIABLES = [
   { id: '{temporary_name}', desc: 'Temporary name if set' },
   { id: '{description}', desc: 'Custom description' },
   { id: '{direction}', desc: 'Wormhole direction (e.g., In, Out)' },
-  { id: '{spawn}', desc: 'Spawn type for outgoing wormholes (e.g., Static, Wandering)' },
+  { id: '{spawn_type}', desc: 'Spawn type for outgoing wormholes (e.g., Static, Wandering)' },
 ];
 
 interface CustomMappingInputProps {
--- a/assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/components/BookmarkNameFormatSetting.tsx
+++ b/assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/components/BookmarkNameFormatSetting.tsx
@@ -7,6 +7,8 @@ import { formatBookmarkName } from '@/hooks/Mapper/helpers/bookmarkFormatHelper'
 import { SignatureGroup, SignatureKind, SystemSignature } from '@/hooks/Mapper/types';
 import { MassState, TimeStatus } from '@/hooks/Mapper/types/connection';
 
+const EMPTY_MAPPING: Record<string, string> = {};
+
 const DUMMY_SIG_BASE: SystemSignature = {
   eve_id: 'ABC-123',
   name: 'ABC-123',
@@ -37,6 +39,8 @@ const VARIABLES = [
   { id: '{mass_status}', desc: 'Mass remaining (e.g., Destab, Crit)' },
   { id: '{temporary_name}', desc: 'Temporary name if set' },
   { id: '{description}', desc: 'Custom description' },
+  { id: '{direction}', desc: 'Wormhole direction (e.g., In, Out)' },
+  { id: '{spawn}', desc: 'Spawn type for outgoing wormholes (e.g., Static, Wandering)' },
 ];
 
 interface CustomMappingInputProps {
@@ -117,6 +121,17 @@ const MASS_OPTIONS = [
 
 const OTHER_OPTIONS = [{ key: 'chain_separator', label: 'Chain Separator', defaultVal: '' }];
 
+const DIRECTION_OPTIONS = [
+  { key: 'direction_outgoing', label: 'Outgoing', defaultVal: 'Out' },
+  { key: 'direction_incoming', label: 'Incoming (K162)', defaultVal: 'In' },
+];
+
+const SPAWN_OPTIONS = [
+  { key: 'spawn_static', label: 'Static', defaultVal: 'Static' },
+  { key: 'spawn_wandering', label: 'Wandering', defaultVal: 'Wandering' },
+  { key: 'spawn_k162', label: 'K162 (Incoming)', defaultVal: 'K162' },
+];
+
 const SIZE_OPTIONS = [
   { key: 'size_small', label: 'Small (Frigate)', defaultVal: 'S' },
   { key: 'size_medium', label: 'Medium', defaultVal: 'M' },
@@ -147,7 +162,7 @@ const CLASS_OPTIONS = [
 export const BookmarkNameFormatSetting = () => {
   const { settings, updateSetting } = useMapSettings();
   const formatStr = settings.bookmark_name_format || '';
-  const customMapping = settings.bookmark_custom_mapping || {};
+  const customMapping = settings.bookmark_custom_mapping || EMPTY_MAPPING;
   const inputRef = useRef<HTMLInputElement>(null);
 
   const [localFormat, setLocalFormat] = useState(formatStr);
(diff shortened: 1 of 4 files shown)


assets/js/hooks/Mapper/helpers/bookmarkFormatHelper.ts
```diff
@@ -361,6 +384,12 @@ export const formatBookmarkName = (
   // Replace {description} -> signature.description
   result = result.replace(/\{description\}/g, () => signature.description || '');
 
+  // Replace {direction} -> incoming (K162) or outgoing (all other types)
+  result = result.replace(/\{direction\}/g, () => getDirectionString(signature.type === 'K162', mapping));
```
guarzo: `{direction}` is decided purely by `signature.type === 'K162'`, so anything that isn't K162 — including a wormhole whose type hasn't been set yet — comes out as "Out".

That case looks reachable to me. `useLinkSignature.ts:44` does `const signatureToLink = { ...signature, group: SignatureGroup.Wormhole }`, so it forces the group on regardless of `type`, and `SystemSignature.type` is a plain `string` that's `''` until someone fills it in. The guard at line 432 only checks `group`, so an unidentified hole sails through to here.

Which is a shame because `getSpawnTypeString` already handles it properly — `if (!signatureType) return '';` on line 44 means `{spawn_type}` goes quiet when it doesn't know. `{direction}` confidently asserts instead, so an unscanned hole that turns out to be a K162 gets bookmarked "Out" and nobody notices until they warp to it.

Making it agree with its neighbour would be a couple of characters — pass `signature.type` in and return `''` when it's empty, same as the other helper:

```ts
const getDirectionString = (signatureType: string | undefined, mapping?: Record<string, string>): string => {
  if (!signatureType) return '';
  if (signatureType === 'K162') {
    return mapping?.direction_incoming !== undefined ? mapping.direction_incoming : 'In';
  }
  return mapping?.direction_outgoing !== undefined ? mapping.direction_outgoing : 'Out';
};
```

Possible you decided "Out" is the better guess for an unknown, in which case ignore me — it's just that the two variables currently disagree about what to do with the same missing information.

assets/js/hooks/Mapper/helpers/bookmarkFormatHelper.ts
```diff
@@ -29,6 +29,28 @@ const getTimeStatusString = (status?: TimeStatus, mapping?: Record<string, strin
...
+const getSpawnTypeString = (
+  signatureType: string | undefined,
+  currentSystemStatics: string[] | undefined,
+  mapping?: Record<string, string>,
+): string => {
+  if (!signatureType) return '';
+  if (signatureType === 'K162') {
+    return mapping?.spawn_k162 !== undefined ? mapping.spawn_k162 : 'K162';
+  }
+  if (currentSystemStatics?.includes(signatureType)) {
```
guarzo: The optional chain here means "statics unknown" and "not one of the statics" produce the same answer. If `getSystemStaticInfo` misses for the current system, `?.statics` is `undefined`, `?.includes(...)` short-circuits, and every wormhole in that system gets labelled "Wandering" — which reads as a definite answer rather than a shrug.

Both call sites pass `getSystemStaticInfo(...)?.statics`, so it depends on the static info being loaded for the system you're standing in. That's very likely true in practice, since the system is on the map. I'd just rather it fell back to `''` than to a confident wrong label:

```ts
if (currentSystemStatics === undefined) return '';
```

Small aside while I'm in this function, not something I'd expect you to fix: a system can roll a wandering hole that happens to share a type with its own static (a C3 with a static D845 picking up a second D845), and there's no way to tell those apart from the sig type alone. Nothing cheap to do about it — just worth knowing "Static" is a best guess rather than a fact in that one case.

assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/components/BookmarkNameFormatSetting.tsx
```diff
@@ -7,6 +7,8 @@ import { formatBookmarkName } from '@/hooks/Mapper/helpers/bookmarkFormatHelper'
 import { SignatureGroup, SignatureKind, SystemSignature } from '@/hooks/Mapper/types';
 import { MassState, TimeStatus } from '@/hooks/Mapper/types/connection';
 
+const EMPTY_MAPPING: Record<string, string> = {};
```
guarzo, line 10: Nice catch on this one — `customMapping` was a fresh `{}` on every render when the setting was unset, and the effect on line 162 has it in the dep array and calls `setLocalMapping` from it, so that's a re-render loop waiting to happen. Hoisting the constant is the right fix.

It is a separate bug from what the PR is about, though, and it's currently invisible unless you know to look for it. Worth a line in the description so it gets its own attention — and it'd probably be a good thing to fix even if the rest of this PR changed shape.

(For what it's worth it looks hard to actually hit today: `DEFAULT_REMOTE_SETTINGS` in `mapRootProvider/constants.ts:10` and the backend default in `map_user_settings_repo.ex:10` both give `bookmark_custom_mapping` a `{}`, so the `||` branch shouldn't normally fire. Still better closed than left open.)

assets/js/hooks/Mapper/helpers/bookmarkFormatHelper.ts
```diff
@@ -394,6 +423,7 @@ export const handleAutoBookmark = async (
   targetSystemClassGroup: string | null,
   targetSystemUuid?: string,
   targetSolarSystemId?: string,
+  currentSystemStatics?: string[],
```
guarzo, line 428: Only a style thought, take it or leave it. `handleAutoBookmark` is up to ten positional parameters now and `formatBookmarkName` eleven, and the preview call in `BookmarkNameFormatSetting.tsx:228` has to pass a literal `undefined` as padding to reach the new one. The new `currentSystemStatics` also lands after the `target*` params, which are about a different system, so it's easy to slot into the wrong slot without anything complaining.

An options object for the tail end (`{ targetSystemClass, targetSystemName, ..., currentSystemStatics }`) would make the call sites self-describing. Definitely not something I'd hold this PR for — more that this is the second parameter added recently and the next one will be worse.

guarzo (commented):
This is a nice little gap-filler. `{direction}` and `{spawn_type}` are exactly the two things you end up typing by hand every time, and doing it through the existing custom-mapping table rather than hardcoding "In"/"Out" means people who name holes differently aren't stuck with your vocabulary. The default values match what most folks already write, too.

Two questions rather than complaints, both about the unknown cases. A wormhole signature that hasn't been identified yet gets `{spawn_type}` right (it renders empty) but `{direction}` still says "Out" — and since linking a signature to a system forces `group: Wormhole` regardless of whether the type is filled in, I think that's reachable on the normal bookmark-then-scan workflow. Second one's smaller: if the statics for the current system aren't in the cache, everything reads "Wandering" with nothing to distinguish it from a real answer. Both inline, and I may well be misreading either.

Also spotted the `EMPTY_MAPPING` change, which looks like an unrelated fix that snuck in — worth calling out in the description so it doesn't get lost. And since it adds two sections to the settings dialog, a screenshot would be great.

DanSylvest:
Hi! @leesolway - i've added some rules for PR. Please read it and update your PR. https://discord.gg/UF6hfHKe9

Regarding your changes - please attach images how it looks like

assets/js/hooks/Mapper/helpers/bookmarkFormatHelper.ts
leesolway: I won't amend in this PR but that does make sense.

assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/components/BookmarkNameFormatSetting.tsx
leesolway: Updated the description as requested

assets/js/hooks/Mapper/helpers/bookmarkFormatHelper.ts
leesolway: Fixed the fallback.

I couldn't think of a good solution to the static issue sadly.

assets/js/hooks/Mapper/helpers/bookmarkFormatHelper.ts
leesolway: I did, but your reasoning is sound. Amended.

PR #646, opened 2026-08-03
feat: share intel between maps

guarzo:
A map admin can choose another map as an intel source, and system names, labels, descriptions, status, comments and structures are then copied onto their map automatically. Fields that come from the source map become read-only on the subscribing map.

## Screenshots

Settings
image

Context
image

@DmitryPopov @DanSylvest


guarzo:
Self-review follow-up — pushed a fix for something this PR had wrong, found while checking it against the snapshot issue dedo1911 raised on #642.

**The two intel-sharing migrations created their FKs with referential actions that the resources never declared.**

| Column | Migration says | `belongs_to` said |
|---|---|---|
| `maps_v1.intel_source_map_id` | `ON DELETE SET NULL` | *(nothing)* |
| `map_system_comments_v1.inherited_from_map_id` | `ON DELETE CASCADE` | *(nothing)* |
| `map_system_structures_v1.inherited_from_map_id` | `ON DELETE CASCADE` | *(nothing)* |

No snapshots were regenerated either, so `mix ash_postgres.generate_migrations` still wanted to add all three columns. The duplicate migration that would have produced is the real problem: since the resources declared no `on_delete`, it would have recreated those FKs **without** the referential actions — silently changing what happens to inherited comments and structures when a source map is deleted.

Fixed by declaring the actions in each `postgres do ... references do` block (matching the convention in `map_access_list` / `map_invite` / `map_ping`) and regenerating the three snapshots. Codegen now reproduces the hand-written migrations exactly — `nilify_all`, `delete_all`, `delete_all` — and a dry run reports no pending changes for these resources.

**Two things I did not change, flagging for your call:**

1. `map_chain_v1` still shows drift, but it is pre-existing on `main` and unrelated to this PR — `20260425000000_add_map_connection_locked_by.exs` added `locked_at` / `locked_by_id` by hand without regenerating the snapshot. Separate one-file PR incoming.

2. The migration timestamps in this PR are `20260209100000` / `20260209100001` — February, while `main` already has applied migrations from March and April. Ecto will still run them since it goes by what is absent from `schema_migrations`, so this is not a correctness bug, but the ordering is misleading and `mix ecto.rollback` would walk them in a surprising order. Happy to renumber them to current timestamps if you'd prefer.

guarzo:
## Proposal

An opt-in intel source link between maps. A map admin nominates another map they have access to as its intel source; when a system appears on the subscriber map, its intel — custom name, labels, description, status, comments and structures — is copied from the source map.

## What it does

- **Source selection** lives in the **Admin Settings tab of the React in-map `MapSettings` dialog** (`assets/js/hooks/Mapper/.../MapSettings/components/AdminSettings.tsx`), which is already gated `{isAdmin && ...}` — not the LiveView `/maps/:slug/settings` modal. It validates access to the nominated map and rejects circular references and chains.
- **Intel-managed fields become read-only on the subscriber map** — the system context menu and the system settings dialog disable the corresponding controls. `disabled` is threaded into `useTagMenu` / `useStatusMenu` / `useLabelsMenu` rather than spread onto their results at the call site, because the tag submenu renders a custom `template` of buttons a call-site `disabled` would never reach.
- **Read-only is per system, not per map.** The init options payload carries `intel_inherited_system_ids` — the systems visible on both maps, which is exactly the set `sync_system/3` copies into. A new `useIsIntelInherited` hook consumes it, so systems the source map has never mapped stay fully editable and the **Sync Intel** menu item only appears where it does something.
- **The client gate is advisory; the server enforces it independently.** `MapSystemsEventHandler` refuses writes to synced fields (`description`, `labels`, `tag`, `temporary_name`, `status`) on inherited systems, and `SystemsImpl.remove_system_comment/4` refuses deletes of inherited comments. Both are reachable over the WebSocket without the UI.
- **Copying is pull-based**: when a system first appears on the subscriber map, and as a one-time backfill when the source is first configured. There's no push propagation, so a **Sync Intel** item in the system context menu pulls current intel for a system on demand.
- Gated behind `WANDERER_INTEL_SHARING_ENABLED` (default `false`), surfaced to the client via `client_env`.

## Chains are refused in both directions

The six system intel fields are copied verbatim into `MapSystem` columns with **no provenance marker** — unlike comments and structures, which carry `inherited_from_map_id`. So in an A → B → C chain, A's intel reaches C even though C's admin was never authorised against A. `validate_source_access/2` only authorises the immediate source and the existing cycle walk only rejects loops, so neither closes this.

`validate_no_chain/2` therefore rejects:

- subscribing to a map that already has a source (`source_is_subscriber`), and
- subscribing a map that is already somebody else's source (`map_is_a_source`).

It's backed by a new `read :intel_subscribers_of` Ash action, which excludes soft-deleted maps and **fails closed** — if the subscriber lookup errors, the change is refused. No schema change: a depth-limited chain with provenance tracking on the six fields is possible later, but it needs a migration and is out of scope here.

## Other fixes in this round

- **Syncs broadcast.** Comments replay the existing per-record `:system_comment_added` / `:system_comment_removed` events; structures use the coarse `:structures_updated` the client already re-fetches on. Previously a sync mutated comments and structures with no broadcast at all.
- **Manual and automatic sync share one entry point** (`SystemsImpl.sync_intel_for_system/3`). The manual path re-implemented the body and called `Impl.broadcast!/3` directly, skipping `:system_metadata_changed`, so webhook and SSE consumers never saw manual syncs.
- **Switching sources clears the previous source's inherited records.** `delete_inherited/3` scopes to the *current* source, so the old map's comments and structures were otherwise unreachable by any later sync.
- **`intel_source_map_id` lives on the `%WandererApp.Map{}` struct.** The handler previously wrote the Ash resource into the map-state cache's `:map` key, replacing the struct that `Impl.update_subscription_settings/2`, `Impl.update_options/2` and `AclsImpl` all pattern-match on. Writes now go through `Map.Server.update_intel_source_map/2` → `Map.update_intel_source_map_id!/2`, mirroring `update_options!/2`.
- **Rejections reach the user.** `set_intel_source_map` had no fallback clause, so a caller without `manage_map` fell through to the generic `handle_ui_event/3` catch-all returning `{:noreply, socket}` — the client's promise never resolved and the dropdown stuck in its optimistic state. There's now a `forbidden` fallback, and the dropdown reverts and toasts on any refusal.
- Migrations renamed to `20260808151833` / `20260808151834`, both after the last migration on `main` (`20260425000000`).
- `console.error` calls dropped; `AdminSettings` reads `clientEnv` from the provider instead of taking it as a prop.
- Copy fix in `Comments.tsx`: "No comments found here" (was ungrammatical). Deliberate, unrelated to intel sharing — say the word and I'll drop it.

## Verification

- `mix ash.codegen --dry-run`: the only pending change is `map_chain_v1`, which is pre-existing drift on `main` fixed by #649. This PR's three snapshots (`maps_v1`, `map_system_comments_v1`, `map_system_structures_v1`) produce no diff.
- Migrations applied, rolled back (`mix ecto.rollback -n 2`) and re-applied cleanly on a scratch database. `\d` confirms the FKs match the resource declarations:
  - `maps_v1.intel_source_map_id` → `maps_v1(id) ON DELETE SET NULL` (`on_delete: :nilify`), indexed
  - `map_system_comments_v1.inherited_from_map_id` → `maps_v1(id) ON DELETE CASCADE` (`on_delete: :delete`), indexed
  - `map_system_structures_v1.inherited_from_map_id` → same, indexed
- `mix test test/unit/map/intel_sync_test.exs test/unit/map/intel_sync_validation_test.exs` — 29 tests, 0 failures (was 15; new coverage for `inherited_solar_system_ids/2`, `inherited_system?/2`, `clear_inherited_from/2`, `intel_subscribers_of/1` incl. soft-deleted exclusion, and the inherited-comment delete refusal).
- Full `mix test` — 955 tests, 14 failures, none in intel-sharing code. 6 are the `MapConnection` `locked_at` / `locked_by_id` duplication bug that #639 fixes, 6 are `AuthControllerTest` "flash not fetched", 1 is `OpenAPIValidationTest`, and 1 (`SlugRecoveryTest`) appeared in only one of two runs. All pre-existing on `main`.
- `mix format --check-formatted` clean; `mix compile --force` produces no new warnings in touched files.
- `tsc --noEmit`: one error in a touched file (`MapSettings.tsx(30,53)` TS6133), confirmed identical on `origin/main`. Prettier applied.

## Known gaps

- **`intel_inherited_system_ids` is init-fresh only.** Options reach the client through `push_map_event("init", ...)` and there is no options-update push event, so the read-only set — like `intel_source_map_id` before it — is computed at map load. Adding a system that the source also has won't flip it read-only until reload. The server-side enforcement has no such staleness, so the worst case is a control that looks editable and is then refused.
- Propagation remains pull-only: edits on the source map after a system is already present don't reach the subscriber until someone triggers **Sync Intel**. Push propagation needs care around fan-out (one source, many subscribers) and is out of scope.

## Dependencies

None — no shared files with my other open PRs. #639 and #649 are referenced above only as the owners of pre-existing failures/drift, not as prerequisites.

Issue #647, opened 2026-08-06
feat: character based third party access to maps


lancetekk:
Wanderer's map API currently authenticates external clients with a single, static, shared secret, public_api_key, stored per map and checked via secure_compare against the caller's bearer token.

That's breaks down for actual corp/alliance usage in two ways:

1. In-game security problem. The key is tied to possession of a string, not to a person or a role. Whoever holds it has full access, indefinitely, regardless of who they are or whether they're even still in the corp.
2. Operational problem. Because the key isn't tied to membership, access doesn't automatically follow it. If a member leaves (or is kicked), the only way to cut off their access is to rotate the key for everyone, which breaks every legitimate integration using that map, and forces the admin to redistribute the new key to every remaining user. There's no way to revoke one person's access without collateral damage to the whole group.

The key is also not exposed in the UI (as of v1.101.3, where i branched off). There's no self-service flow for a corp member to get acess to a map they already have access to.

Wanderer has a permission model, ACLs with owner/character/corp/alliance roles, but the API boundary doesn't use it. It falls back to the secret.

**Idea:**
Replace "prove you know this secret" with "prove who you are, and let the existing ACL model decide."

The client sends its own EVE SSO access token.

1. Verifies that token against EVE SSO's own verify endpoint, so identity is anchored in CCP's auth.
2. Resolves the corresponding character, and from there corp/alliance membership.
3. Runs that identity through the map's existing ACL permission logic.
4. On success, mints a short-lived (15 min), map-scoped token encoding exactly the permission mask that specific character is entitled to.

The static key keeps working exactly as before. Rhe new token type is tried first and is structurally distinguishable, so nothing depending on the old key should break.

The API now knows which character is actually calling, not just "someone with the key," per-user personalization becomes possible at the API level, a caller can now get their own configured route hubs (/user-routes) instead of everyone getting the map owner's.

**Is this useful?**
Yes. I just need it to display short WH connections on Homeassistant, but others might use it for more important stuff.

PR #648, opened 2026-08-06
 #647: character based access to maps

lancetekk:
This implements #647 

The required changes have been implemented and tested locally in a container. The endpoint behaves as expected.
I tested it with https to make sure the callbacks work properly, i left the required config infrastructure in there for manual testing.

The guidelines in https://github.com/wanderer-industries/wanderer/wiki are mostly related to the frontend, the PR does not touch the JS part.

Disclaimer: i am not familiar with elixir and therefore had Claude help me with the code production.


lib/wanderer_app_web/controllers/plugs/check_map_api_key.ex
```diff
@@ -4,16 +4,51 @@ defmodule WandererAppWeb.Plugs.CheckMapApiKey do
...
+        Logger.warning("Missing or invalid 'Bearer' token")
+        conn |> respond(401, "Missing or invalid 'Bearer' token") |> halt()
+
+      [_non_bearer_token] ->
+        Logger.warning("Invalid authorization format - Bearer token required")
+        conn |> respond(401, "Invalid authorization format - Bearer token required") |> halt()
+    end
+  end
+
+  defp call_with_eve_token(conn, %{map_id: map_id, character_eve_id: eve_id}) do
```
guarzo, line 36: Two things land right here.

The big one, covered in my main comment: this assigns `:current_character` the same way the static-key branch does, but the two aren't equivalent any more. With the static key, `current_character` is always the map owner, so "got past this plug" and "may do anything" were the same statement. Now a viewer can get past it, and nothing downstream checks.

The smaller one: `map_id` comes purely from the token and `conn.params["map_identifier"]` is never consulted (the static branch does, via `fetch_map_id/1`). So a token for map A sent to `/api/maps/B/systems` quietly returns map A's systems with a 200. Not an escalation, since you do have access to A, but it makes the path param meaningless. Worth comparing the two and 403-ing on a mismatch. #635 does exactly this in `MapSystemsConnectionsController`, normalizing both sides with `to_string/1`.

lib/wanderer_app/map/eve_token_auth.ex
```diff
@@ -0,0 +1,136 @@
...
+  @spec exchange(String.t(), String.t()) ::
+          {:ok, map()}
+          | {:error,
+             :map_not_found | :invalid_token | :verify_failed | :character_not_tracked | :forbidden}
+  def exchange(map_identifier, eve_access_token) do
+    with {:ok, map} <- resolve_map(map_identifier),
+         {:ok, eve_character} <- @eve_sso.verify_token(eve_access_token),
+         {:ok, character} <- resolve_character(eve_character),
+         {:ok, map} <- load_acls(map),
+         mask when mask not in [0, -1] <- permission_mask(map, character) do
```
guarzo, line 49: `:viewer` is mask 7, so it sails straight through this guard and comes out the other side with a token that's indistinguishable from the map owner's once it reaches the plug. The `permissions` map you build a few lines down never makes it into the signed payload, so nothing downstream could act on it even if it wanted to.

If you go the narrow route for now, this is the line to tighten, something along the lines of requiring `check_permission(mask, admin_map)`.

lib/wanderer_app_web/controllers/map_api_controller.ex
```diff
@@ -814,6 +851,159 @@ defmodule WandererAppWeb.MapAPIController do
...
+        conn |> put_status(:bad_request) |> json(%{error: msg})
+
+      {:error, reason} ->
+        conn
+        |> put_status(:internal_server_error)
+        |> json(%{error: "Could not compute user routes: #{APIUtils.format_error(reason)}"})
+    end
+  end
+
+  defp fetch_required_param(params, key) do
```
guarzo, line 951: Two small ones here.

`APIUtils.require_param/2` already does this, and you're using it over in `EveAuthController`, so this can probably go.

Also the value only gets checked for "non-empty string", but it eventually hits `String.to_integer/1` in `Routes.do_find_routes/4`. So `?system_id=abc` raises and surfaces as a 500 rather than a 400.

lib/wanderer_app_web/controllers/map_api_controller.ex
```diff
@@ -814,6 +851,159 @@ defmodule WandererAppWeb.MapAPIController do
...
+      {:error, msg} -> {:error, :missing_param, msg}
+    end
+  end
+
+  defp maybe_put_string(settings, _key, nil), do: settings
+  defp maybe_put_string(settings, key, value), do: Map.put(settings, key, value)
+
+  defp maybe_put_avoid(settings, nil), do: settings
+
+  defp maybe_put_avoid(settings, value) do
```
guarzo, line 998: This produces a list of strings, but it gets merged into an avoidance list of integers over in `map_routes.ex:198` (`@default_avoid_systems` plus the edencom/trig/pochven ids) before being posted to the route service. So the `avoid` param quietly does nothing. Wants `String.to_integer` with a sensible error for junk input.

config/dev.exs
```diff
@@ -20,6 +20,13 @@ config :wanderer_app, WandererAppWeb.Endpoint,
   # Binding to loopback ipv4 address prevents access from other machines.
   # Change to `ip: {0, 0, 0, 0}` to allow access from other machines.
   http: [ip: {0, 0, 0, 0}, port: 4444],
+  https: [
```
guarzo, line 23: I think this one wants to come out. You mentioned leaving the config in for manual testing, but `priv/cert/` is gitignored in this PR and Cowboy won't start the https listener when the keyfile is missing, so `mix phx.server` stops booting for anyone who hasn't run `mix phx.gen.cert`. There's already a commented-out version of this further down the file (~line 50), which is presumably why it's commented out.

.devcontainer/docker-compose.yml
```diff
@@ -13,6 +13,8 @@ services:
       - db-new:/var/lib/postgresql/data
 
   wanderer:
+    env_file:
+      - .env
```
guarzo, line 17: `.env` is gitignored (`.gitignore:34`) and compose hard-errors when a declared `env_file` is missing, so this breaks the devcontainer for anyone who doesn't happen to have that file. Either drop it, or mark it `required: false` (needs compose spec 2.24+).

guarzo (commented):
Had a proper read through this. The core idea is good, and reusing `check_characters_access/2` so API access follows the same ACL logic as the web UI is exactly the right instinct. Nothing here is a start-over. It's mostly one authorization gap plus some dev-env bits that hitched a ride.

(Also, no need for the disclaimer in the description. Plenty of us lean on Claude (just avoid it for front-end work, or @DanSylvest will be annoyed). The code reads fine.)

## The thing that needs sorting first

**The minted token grants full map-owner powers regardless of the character's role.** `exchange/2` works out the permission mask, puts it in the JSON response, and then that's it. It never goes into the signed token and nothing downstream ever checks it. Meanwhile `call_with_eve_token/2` assigns `:current_character` exactly like the static-key branch does, and that branch is only safe because holding `public_api_key` *is* the authorization decision, so `current_character` is always the map owner.

Nothing under `:api_map` re-checks anything:

```
$ git grep -n "check_permission\|Permissions\." -- lib/wanderer_app_web/controllers/
(nothing)
```

So a `viewer` (mask 7, sails past the `not in [0, -1]` check) can mint a token and then hit `DELETE /systems`, `POST /acls`, `POST /duplicate`, `PUT /webhooks/toggle`, basically the whole write surface. `MapAPIController.check_map_owner/2` is the only guard anywhere in the controllers, and it exists precisely because everything else could assume authenticated == owner.

Two ways out, whichever suits you:

- Put the mask in the signed payload, assign it, and gate the write routes on `Permissions.check_permission/2`. A small plug taking the required bit as `opts` would slot into the existing pipeline nicely.
- Or scope the PR down and only mint tokens for admin-level masks, with a note in the moduledoc that role support is a follow-up. Less useful, but honest and safe.

## A couple of general notes

**Worth a look at #635.** Different surface (`/api/v1`, enforced through Ash policies), so no overlap or conflict with this. But its `MapSystemsConnectionsController` already solves the token-map-vs-path-map problem I've flagged inline, `to_string/1` normalization and all, with a regression test that an upper-cased own-map id still 404s. Free pattern to lift.

Heads-up on that same PR though: it adds deny-by-default policies to `Api.Map`, and this PR's `MapApi.by_id/1` and `Ash.load(map, acls: ...)` calls pass no actor. That's fine today, because #635 also sets `authorize :when_requested` on the domain so actor-less calls skip policies entirely. Just flagging the coupling, since nothing here would fail loudly if that setting ever changed.

**The two tests I'd most want to see** are a viewer token being turned away from a write endpoint, and a map-A token being turned away from map B. Both would fail today.

**On `test/integration/api/map_user_routes_controller_test.exs`:** `Routes.find` bottoms out in `Esi.get_routes_custom`, which POSTs to a live external routing service (`esi/api_client.ex:60`) with no config indirection to mock out. Without network it falls through to `get_routes_eve`, which returns stubs with `success: false` and dumps a JSON file into tmp, so the `destination` assertion may well be passing for the wrong reason. Probably wants the same swappable-module trick you used for `:eve_sso`, which was nicely done by the way.

Smaller: the two new test files are missing `async: false`, which everything else in `test/integration/api/` has.

Rest is inline. Happy to sketch out the permission-gating plug if that'd help.

lancetekk:
Thank you so much for the detailed response. I will take care of the mentioned points and report back if i have questions and/or a resolution next week!

lancetekk:
I tried adressing all the comments one by one. I had a few issues when mapping the permissions from the mask to the endpoints, but i hope it works as intended. I did test it locally, but since it worked before, it also works now. 

There is a handful of tests failing, but from my limited understanding it does not seem like i broke them.

lancetekk:
I tried adding a permission-gating test for rotate_secret.The test factory's `create_map_webhook_subscription/1` helper  failed immediately with `NoSuchAttribute: :secret` when building the changeset
Webhook creation appears to have an issue.

`MapWebhookSubscription's :create` action generates the secret via Ash.Changeset.force_change_attribute(changeset, :secret, secret) (map_webhook_subscription.ex:141). :secret is an AshCloak-encrypted attribute, and` force_change_attribute` raises NoSuchAttribute for it, therefore my webhook creation (POST /webhooks, or a direct Ash.create/3) failed.

Since webhooks_enabled defaults to false and there was no test coverage on this controller, this seems to have gone unnoticed or i am just massively mistaken.

PR #649, opened 2026-08-08
fix(db): add the missing foreign key on connection locks

guarzo:
The column that records who locked a connection has no foreign key, so deleting a character can leave connections pointing at a character that no longer exists. This adds the constraint and clears any stale values.

@DmitryPopov @DanSylvest

### DIFF
--- a/lib/wanderer_app/api/map_connection.ex
+++ b/lib/wanderer_app/api/map_connection.ex
@@ -15,6 +15,12 @@ defmodule WandererApp.Api.MapConnection do
       # Critical index for list_connections query performance
       index [:map_id], name: "map_chain_v1_map_id_index"
     end
+
+    references do
+      # A deleted character releases its locks rather than taking the
+      # connections with it -- the connection outlives whoever locked it.
+      reference :locked_by, on_delete: :nilify
+    end
   end
 
   json_api do
--- a/priv/repo/migrations/20260808152615_add_map_connection_locked_by_fk.exs
+++ b/priv/repo/migrations/20260808152615_add_map_connection_locked_by_fk.exs
@@ -1,0 +1,48 @@
+defmodule WandererApp.Repo.Migrations.AddMapConnectionLockedByFk do
+  @moduledoc """
+  Adds the missing foreign key on `map_chain_v1.locked_by_id`.
+
+  20260425000000 added the column as a bare `:binary_id` with no constraint,
+  while `MapConnection` declares `belongs_to :locked_by, Character`. The
+  resource therefore promised a relationship the database never enforced, and
+  because no snapshot was regenerated at the time, every subsequent
+  `mix ash_postgres.generate_migrations` run wanted to re-add both
+  `locked_by_id` and `locked_at` from scratch.
+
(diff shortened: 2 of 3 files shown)


guarzo:
## Proposal

`map_chain_v1.locked_by_id` has no foreign key, and its resource snapshot is stale.

## The problem

`20260425000000_add_map_connection_locked_by.exs` added the column as a bare `:binary_id`:

```elixir
add :locked_by_id, :binary_id
```

while `MapConnection` declares `belongs_to :locked_by, WandererApp.Api.Character`. So:

1. **The resource declares a relationship the database never enforced.** Ash treats `locked_by_id` as a reference to `Character`; Postgres treats it as an unconstrained `binary_id`. `Character` has a plain `:destroy` action, so nothing at the database level stops a delete from leaving a connection pointing at an id that is no longer there. This PR closes the gap between what the resource declares and what the schema enforces — it is not filed on the back of an observed incident, and I have not gone looking for existing bad rows in production.
2. **No snapshot was regenerated**, so `mix ash_postgres.generate_migrations` has wanted `locked_at` and `locked_by_id` from scratch since April. It emits `add` for both — which fails with a duplicate-column error on every database the April migration already ran on. Codegen output can't be used as-is here; the migration has to be hand-shaped as an `alter`.

## The fix

- Adds the constraint with `ON DELETE SET NULL`. The columns already exist, so `up/0` only alters — it never re-adds them.
- Declares the referential action on the resource via a `references` block, matching the convention in `map_access_list` / `map_invite` / `map_ping`. A deleted character releases its locks rather than taking the connections with it.
- `up/0` nullifies any `locked_by_id` with no matching character before adding the constraint. This is belt-and-braces: such rows are possible given (1), and a single one would abort the `ALTER`, so the migration should not depend on their absence. On a database that has none, the statement is a no-op.
- `down/0` is deliberately asymmetric — it drops only the constraint. The columns predate this migration, so removing them on rollback would destroy data it never created. Noted in a comment.
- Regenerates the `map_chain_v1` snapshot, so the next `mix ash.codegen` is quiet.

## Verification

Against a scratch database built from the full migration history:

| Check | Result |
|---|---|
| `up` applies | ✅ |
| Constraint present with `confdeltype = 'n'` (SET NULL) | ✅ |
| `down` drops constraint, leaves both columns and their rows | ✅ |
| Re-apply after rollback | ✅ |
| Row seeded with an orphaned `locked_by_id` | row survives, lock released, migration succeeds |
| `mix ash.codegen --check` | no pending changes anywhere in the project |
| `ADD CONSTRAINT ... NOT VALID`? No — plain `ADD CONSTRAINT` | `relfilenode` unchanged (19314 → 19314): validating scan, no table rewrite |

The seeded-orphan row is the one that matters: it is what makes the nullify step verifiable rather than merely defensive. The last row is there because "adding an FK rewrites the table" is a common assumption — it does not; it takes a `SHARE ROW EXCLUSIVE` lock and scans to validate.

## Dependencies

None — no shared files with my other open PRs. #650 adds a CI check that reports exactly this drift; it's independent and can land in either order.

PR #650, opened 2026-08-08
ci: report when Ash resources and their snapshots disagree

guarzo:
Resource definitions can drift from their stored snapshots without anything failing, which later shows up as a surprise or colliding migration. CI now reports the drift in the test results comment; it does not block merging.

@DmitryPopov @DanSylvest

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,6 +72,22 @@ jobs:
           fi
         continue-on-error: true
 
+      # Advisory, like the other checks here. Catches resource definitions that
+      # have drifted from priv/resource_snapshots — usually after a dependency
+      # bump changes how a default is serialized, which is otherwise invisible
+      # until someone runs `mix ash.codegen` locally and gets an unrelated diff.
+      - name: Check Ash codegen is up to date
+        id: codegen
+        run: |
+          if mix ash.codegen --check; then
+            echo "status=✅ Up to date" >> $GITHUB_OUTPUT
+            echo "count=0" >> $GITHUB_OUTPUT
+          else
+            echo "status=❌ Pending changes" >> $GITHUB_OUTPUT
+            echo "count=1" >> $GITHUB_OUTPUT
+          fi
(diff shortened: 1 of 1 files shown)


guarzo:
## Proposal

Adds an advisory `mix ash.codegen --check` step to the test workflow, plus a row in the results table.

Resource definitions can drift from `priv/resource_snapshots/` without anything failing. The drift stays invisible until someone runs `mix ash.codegen` locally for an unrelated change and gets a surprise diff — or worse, generates a migration that collides with a hand-written one already in `priv/repo/migrations/`.

`continue-on-error: true`, matching the other checks in this job: `main` currently has drift on `map_chain_v1`, so this reports rather than gates.

## Dependencies

None. The `map_chain_v1` drift this check reports is fixed by #649; the two no longer touch any of the same files.


## Review pass

Two changes since the first push:

- The codegen step now runs **after** the compile step. Ordered before it, a branch that simply doesn't build reported "codegen drift", which points reviewers at the wrong thing.
- The drift itself is surfaced in the PR comment (last 40 lines of the codegen output, as a multiline `GITHUB_OUTPUT` block) rather than only in the job log. A row that says "drift" with no detail costs a click and a scroll to act on.

Two things worth flagging rather than fixing:

- **The codegen row is deliberately excluded from the score.** It cannot pass on `main` today, so counting it would make every PR's score look one worse for a pre-existing condition.
- **`actionlint` reports 14 more findings after this change** (SC2086 42 → 54, SC2129 3 → 5). All of them come from `echo … >> $GITHUB_OUTPUT` lines. Both classes already occur 45 times in this file, so quoting only the new ones would leave the file inconsistent with itself; I've matched the surrounding convention instead. Pre-existing `SC2034`, `SC2320`, and the `actions/cache@v3`-too-old finding are unchanged by this PR.

## Overlap with #642

#642 also edits `.github/workflows/test.yml`. The two touch different steps and I expect them to merge cleanly, but whichever lands second should be re-checked rather than assumed.

PR #651, opened 2026-08-08
refactor(map): move the wormhole class list into one module

guarzo:
The list of wormhole system classes was written inline in the connection code, and a second feature needs the same list. It now lives in one module with helpers, and behaviour is unchanged.

First of the stack #651 → #652 → #653 → #654, but useful on its own.

@DmitryPopov @DanSylvest

### DIFF
--- a/lib/wanderer_app/map/server/map_server_connections_impl.ex
+++ b/lib/wanderer_app/map/server/map_server_connections_impl.ex
@@ -19,13 +19,9 @@ defmodule WandererApp.Map.Server.ConnectionsImpl do
   @ns 9
   # @ccp2 10
   # @ccp3 11
-  @thera 12
+  # Classes 12 and 14-18 (Thera and the drifter systems) are not named here:
+  # WandererApp.SystemClass owns the canonical wormhole class list.
   @c13 13
-  @sentinel 14
-  @barbican 15
-  @vidette 16
-  @conflux 17
-  @redoubt 18
   @a1 19
   @a2 20
   @a3 21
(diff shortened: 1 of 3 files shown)


guarzo:
## Proposal

Extracts the wormhole system-class list out of `map_server_connections_impl.ex` into a single `WandererApp.SystemClass` module.

## What it does

- New `lib/wanderer_app/system_class.ex` — the canonical class list plus `wormhole_classes/0`, `wormhole?/1` (class → bool) and `wormhole_system?/1` (solar system id → bool, via `CachedInfo`). There is no class-name lookup; an earlier draft of this description claimed one.
- `map_server_connections_impl.ex` calls it instead of matching on an inline list. Behavior is unchanged; the same ids are covered.
- New `test/unit/system_class_test.exs` pins the membership so a future edit to the list can't silently change connection classification.

The list was inline in one module and needed by a second (Discord kill notifications, #653) to decide whether a kill happened in wormhole space.

## Dependencies

None. This is the base of a four-PR stack (#651 → #652 → #653 → #654) but stands alone and is useful on its own.

## Verification

- `mix test test/unit/system_class_test.exs` — 9 tests, 0 failures
- `mix credo --strict` — no findings in the new module
- `mix format --check-formatted` on the changed files

`mix format --check-formatted` is currently failing on `main` for files this PR doesn't touch (`router.ex`, `k162_backlink_preference_test.exs`, `openapi_spec_analyzer.ex`, two migrations). Pre-existing, left alone.


## Review pass

`wormhole_system?/1` now has an `is_integer/1` guard and a logging catch-all clause instead of falling through to `CachedInfo` with whatever it was handed:

```elixir
def wormhole_system?(solar_system_id) when is_integer(solar_system_id) do
  # ... resolve via CachedInfo, log and return false if it can't be resolved
end

def wormhole_system?(solar_system_id) do
  Logger.warning("[SystemClass] wormhole_system?/1 expects an integer solar system id, got: #{inspect(solar_system_id)}")
  false
end
```

This matters because of who calls it: `DiscordDispatcher` (#653) passes `system_id` straight off a zkillboard payload, where a missing or malformed field arrives as `nil` or a string. The `@spec` widened to `term()` to match — a narrower `integer()` spec alongside a catch-all clause is a dialyzer contract mismatch, not documentation.

The test module also moved from `ExUnit.Case` to `WandererApp.DataCase` (still `async: true` — `DataCase` supports it) since the resolver path touches the repo.

## Who uses this

Two call sites across the stack, which is the argument for the module existing at all:

| caller | uses |
|---|---|
| `map/server/map_server_connections_impl.ex:41` | `wormhole_classes/0` — compile-time list for the existing connection classification |
| `external_events/discord_dispatcher.ex:305` | `wormhole_system?/1` — the `wh_only` notification filter (#653) |

The second is why `wormhole_system?/1` is here rather than in #653: it is the same question the connection code already asks, and it should have one answer.

PR #652, opened 2026-08-08
feat(discord): add the table that stores a map's Discord webhook

guarzo:
Adds the database table and resource that hold a map's Discord webhook settings, including the encrypted URL. Nothing uses it yet — delivery is #653.

Merge after #651.

@DmitryPopov @DanSylvest


guarzo:
## Proposal

Adds the `MapDiscordNotification` Ash resource, its migration, and its snapshot. Nothing consumes it yet — the delivery pipeline is #653.

## What it does

- `lib/wanderer_app/api/map_discord_notification.ex` — one row per map: webhook url, `enabled?`, `wh_only`, plus delivery bookkeeping (`last_delivery_at`, `last_error`, `consecutive_failures`).
- The webhook url is an AshCloak encrypted attribute and is deliberately **not** `decrypt_by_default` — callers needing the plaintext load it explicitly, so it doesn't ride along in query results or error reports.
- `priv/repo/migrations/20260801234058_add_map_discord_notifications.exs` and the matching snapshot, both generated by `mix ash.codegen`. The resource's `references` block declares `on_delete: :delete` so the resource and migration agree.
- `test/unit/api/map_discord_notification_test.exs` — 28 tests, mostly url validation.

## Where the decisions are

`valid_webhook_url?/1` enforces https, an allowlist of Discord hosts, and the `/api/webhooks/<id>/<token>` path shape. Two specifics worth a look:

- **Control characters are rejected rather than stripped.** `URI.parse/1` keeps them inside the path, so `.../api/webhooks/1/abc\ndef` clears the segment checks and would reach Finch as a request URL.
- **Surrounding whitespace is tolerated, not rejected.** Ash's `:string` type trims before storing, so a url pasted with a trailing newline is stored clean. Rejecting it here would make this function — which the settings form also calls to pre-validate raw input — disagree with what the write actually does. A test asserts the trim rather than assuming it, so if that Ash default changes we find out here instead of at delivery time.

## Dependencies

Requires #651. Cross-fork PRs can't base on a fork branch, so this targets `main` and the diff shrinks once #651 merges — review `lib/wanderer_app/api/map_discord_notification.ex`, the migration, the snapshot, and the resource test.

## Verification

- `mix test test/unit/api/map_discord_notification_test.exs` — 28 tests, 0 failures
- `mix ash.codegen --check` — no drift for this resource
- `mix credo --strict` — no findings in the new module

`mix ash.codegen --check` does report pending drift for `map_chain_v1`. That drift is present on `main`, untouched by this branch, and is addressed by #649.


## Review pass

**Snapshot filename now matches its migration.** The snapshot was written as `20260801234059.json` while the migration is `20260801234058_add_map_discord_notifications.exs` — one second apart, which is what you get when codegen stamps the two files across a tick boundary. Renamed to `20260801234058.json` so the pairing is greppable, since that off-by-one is exactly the kind of thing that makes the next person's snapshot archaeology take an hour.

**`exclude_inputs: [:webhook_url]` on three `code_interface` defines.** AshCloak adds a `webhook_url` *argument* to every action on the resource regardless of what that action `accept`s. Without the exclusion, `record_success/1`, `record_failure/2` and `disable/2` all advertise a `webhook_url` input they have no business taking — and a caller passing one would re-encrypt the field as a side effect of recording a delivery result:

```elixir
define(:record_success, action: :record_success, exclude_inputs: [:webhook_url])

define(:record_failure,
  action: :record_failure,
  args: [:error],
  exclude_inputs: [:webhook_url]
)

define(:disable, action: :disable, args: [:error], exclude_inputs: [:webhook_url])
```

This is a property of AshCloak rather than of these actions, so it's worth knowing if you add a fourth.

PR #653, opened 2026-08-08
feat(discord): send a map's kills to its Discord webhook

guarzo:
Sends kills that happen on a map to the Discord webhook configured for that map, with retries and back-off when Discord is slow or rate limits us. The settings screen to configure it is #654.

Merge after #652 (which needs #651).

@DmitryPopov @DanSylvest


guarzo:
## Proposal

The delivery pipeline for per-map Discord kill notifications: a per-map worker, the embed formatter, the HTTP client, and the dispatcher that routes killmail events into them.

This is the design-heavy PR of the four — everything with a retry policy, a failure classification, or a supervision decision is here.

## Shape

`MapEventRelay` → `DiscordDispatcher` (cached config lookup + formatting) → `WorkerSupervisor.deliver` → `Worker.enqueue` (a `GenServer.cast`) → HTTP.

The dispatcher does no I/O — it reads the per-map config from a Cachex-backed cache, formats embeds, and casts, so HTTP delivery is off the dispatcher's process. Same shape as the existing `WebhookDispatcher.handle_cast` → `process_webhook_delivery` path.

One worker per map, registered in a Registry, started on demand by a DynamicSupervisor, shut down after 60s idle. Delivery uses an isolated Finch pool (`:discord_finch`) so a slow Discord can't consume connections from anything else.

## Decisions worth arguing about

- **Failure classification.** 429 retries honoring `retry-after`, but only when it fits the event's remaining budget; a `retry-after` beyond the budget fails the event outright rather than retrying early (see the review-pass note below). 5xx and transport errors retry with capped exponential backoff, 5 attempts, 60s per-event deadline. 4xx doesn't retry but counts as a failure. 404 disables the notification immediately — the webhook is gone, not failing. 10 consecutive failed events disables it. An unencodable payload is classified permanent: retrying can't make it encodable, and doing so would burn the attempt budget and surface to the user as a "connection error."
- **Per-event success semantics.** A multi-chunk event that succeeds on chunk 1 and fails on chunk 2 records a failure and does *not* stamp `last_delivery_at` — otherwise a partially-delivered event looks clean in the settings tab.
- **Queue bound.** 100 events per worker, dropping oldest. A map under a heavy fight shouldn't grow a worker mailbox without limit, and recent kills are the ones worth keeping.
- **Kill-switch tolerance.** `application.ex` starts the Discord tree only when webhooks are enabled, so with the feature off the registry doesn't exist. `WorkerSupervisor.deliver/3` and `stop_worker/1` return rather than raise in that case, and the dispatcher's cast to an unregistered name is a no-op. Covered by a test.
- **Events with no `solar_system_id` are dropped explicitly.** With `wh_only` on they'd be dropped anyway but for the misleading reason "not wormhole space"; with `wh_only` off they'd reach the formatter and render an embed for a nil system.
- **A disabled notification stays disabled when its url is replaced.** Re-enabling on url change would override an explicit `enabled?: false` arriving in the same form submission, resurrecting a config the failure threshold deliberately disabled. Easy to change if you'd rather it auto-recover.

`WorkerSupervisor` starts before `DiscordDispatcher` so the first event doesn't find the worker tree missing; both sit inside the same `if webhooks_enabled` branch as the existing `WebhookDispatcher`.

## Config

Two new env vars, in `config/runtime.exs` only, and added to `.env.example`:

```
WANDERER_FINCH_DISCORD_POOL_SIZE   (default 10)
WANDERER_FINCH_DISCORD_POOL_COUNT  (default 1)
```

## Dependencies

Requires #652 (resource), which requires #651. Cross-fork PRs can't base on a fork branch, so this targets `main` and the diff is cumulative until the parents merge — the files to review here are the six under `lib/wanderer_app/external_events/` plus `application.ex` and `config/runtime.exs`.

## Verification

- `mix test test/unit/external_events/` — 93 tests across the stack, 0 failures
- Worker tests cover retry-after honoring, non-blocking retry (a long `retry-after` must not stall the mailbox), queue overflow, 403/401/404 handling, the 10-failure disable threshold, per-event success semantics, deadline expiry, idle shutdown, and the supervisor-down path
- `mix credo --strict` — no findings in any new module
- `mix format --check-formatted` on all changed files


## Review pass

**The 429 clamp is gone, and that changes behavior you should look at.**

`parse_retry_after/1` previously clamped with `min/2`, so a `retry-after` of 300s became a 10s retry — i.e. we ignored what Discord told us and hammered it early, which is the thing rate limiting exists to prevent. Now an over-budget `retry-after` fails the event:

```elixir
defp handle_rate_limited(state, headers) do
  case retry_after_ms(headers) do
    ms when ms > @retry_after_give_up_ms ->
      finish(state, {:error, "Discord returned 429 with retry-after #{ms}ms exceeding budget", :count})

    ms ->
      schedule_retry(state, ms, "Discord returned 429 (rate limited)")
  end
end
```

The `:count` tag is the part worth arguing about: these failures count toward `@max_consecutive_failures`, so **a map that sits behind a long Discord rate limit for 10 consecutive events now gets its notification disabled**, where before it would have kept retrying early forever. I think that's the right trade — a persistently over-budget webhook is not delivering, and silently retrying past the limit is worse than telling the user — but it is a real behavior change and the alternative (tag it `:ignore` so rate limiting never disables anything) is a one-word edit if you disagree.

**Map soft delete now runs the notification cleanup.** `Map`'s `:mark_as_deleted` action, not the hard `:destroy`, is the path users actually take (`maps_live.ex`, `admin_maps_live.ex`), so it now carries the same `after_transaction(&after_destroy/3)` hook. Without it, deleting a map left its Discord workers and subscription rows behind.

**Finch pool config moved to match convention.** The Discord pool settings lived in their own `config :wanderer_app, :discord_finch` block with a single `WANDERER_DISCORD_POOL_SIZE` var. They now sit in the main `config :wanderer_app` block alongside `finch_webhooks_*` / `finch_default_*`, and follow the same naming and size/count pairing:

| before | after |
|---|---|
| `WANDERER_DISCORD_POOL_SIZE` (default 10) | `WANDERER_FINCH_DISCORD_POOL_SIZE` (default 10) |
| — | `WANDERER_FINCH_DISCORD_POOL_COUNT` (default 1) |

The old var was functional — it was read by `discord_pool_size/0` — so this is a rename for consistency, not a bug fix. It has not shipped, so there is nothing to deprecate.

PR #654, opened 2026-08-08
feat(discord): add a Notifications tab to the map settings page

guarzo:
Map owners get a Notifications tab where they set the Discord webhook, limit kills to wormhole space, exclude systems, send a test message and see the delivery status. Last of the four-PR Discord stack.

Merge after #653.

## Screenshots

Setup — the Notifications tab:

setup

A delivered notification in Discord:

notification

@DmitryPopov @DanSylvest


guarzo:
## Proposal

Adds the Notifications tab on the map settings page, where a map owner configures Discord kill notifications, and the news post documenting the feature.

## What it does

- `map_notifications_component.ex` — a LiveComponent for the webhook url, the enable toggle, the wormhole-only toggle, a test-delivery button, and current delivery status (last delivery, last error, consecutive failures).
- The tab sits in the same tab strip as the existing **Bot** tab on the map settings page, gated by `@webhooks_enabled?` and the same map-owner check the neighbouring tabs use. This is the LiveView settings page, not the in-map React "Admin Settings" dialog.
- `core_components.ex` — `live_select/1` gains optional `id` and `phx-target` attrs, forwarded only when non-nil, so existing callers are unaffected.
- Systems can be excluded individually, through a search picker backed by `CachedInfo` rather than a database read per render.

## Dependencies

Requires #653 (delivery pipeline), which requires #652 (resource), which requires #651. Cross-fork PRs can't base on a fork branch, so this targets `main` and the diff is cumulative until the parents merge — the files to review here are `map_notifications_component.ex`, the `maps_live` wiring, and `core_components.ex`.

## Verification

- `mix test test/wanderer_app_web/live/map_notifications_test.exs` — covers the owner gate, url validation feedback, toggle persistence, test delivery, and status rendering
- Full stack suite: 93 tests, 0 failures
- `mix credo --strict` — no findings in the new modules
- No debug artifacts; zero `.tsx` files touched

## Known gap

The tab strip has no ARIA roles or tabindex handling. Adding them means changing the existing Public Api and Bots tabs too, so it's left for a follow-up rather than done partially here.


## Review pass

- **`MapNotificationsComponent` moved into the `Maps` namespace** — `WandererAppWeb.MapNotificationsComponent` → `WandererAppWeb.Maps.MapNotificationsComponent`, matching its path (`live/maps/components/`) and its siblings. Callers updated in `maps_live.ex` and `maps_live.html.heex`.
- **The masked-token prefix length is now a named attribute** (`@masked_url_token_prefix 4`) instead of a bare literal in the masking helper.
- **Dropped the `by_solar_system_ids` read action** from `MapSolarSystem` (and its `code_interface` define). It existed only to label the excluded-systems chips; that now goes through `CachedInfo.get_system_static_info/1`, which is already warm, so the label helper no longer issues a database read per render:

  ```elixir
  defp excluded_system_labels(%{excluded_systems: ids}) do
    Enum.map(ids, fn id ->
      label =
        case CachedInfo.get_system_static_info(id) do
          {:ok, %{solar_system_name: name, region_name: region}} -> "#{name} (#{region})"
          _ -> to_string(id)
        end

      {id, label}
    end)
  end
  ```

  Removing a *read* action needs no snapshot regeneration — Ash resource snapshots record attributes, identities, indexes and constraints, but carry no `actions` key at all. `mix ash.codegen --check` is quiet after this change.

PR #656, opened 2026-08-11
Map export and import

z4nr34l:
Split out of #632 at @DmitryPopov's request - this is that branch's export/import commit on its own, nothing else changed.

Duplication copies a map inside one instance but has no serialized form, so chains cannot be backed up or moved between deployments.

`WandererApp.Map.Operations.Transfer` produces a versioned document with systems, connections and signatures, and replays it into an existing map through the map server, so connected clients see systems appear as they are added.

- Available from the Import/Export tab of the map user settings, and over the API: `GET /api/maps/{map_identifier}/export`, `POST /api/maps/{map_identifier}/import`.
- Import adds what is missing and leaves systems already on the map untouched, so it is safe to run against a live chain. It is deliberately not a destructive restore.
- Access lists, subscriptions and user settings stay out of the document - they reference accounts and characters that do not exist on the importing side.
- Signature import upserts on `(system_id, eve_id)`, so re-importing does not duplicate signatures.

### DIFF
--- a/assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/components/ImportExport.tsx
+++ b/assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/components/ImportExport.tsx
@@ -1,5 +1,5 @@
 import { useMapRootState } from '@/hooks/Mapper/mapRootProvider';
-import { useCallback, useMemo, useRef } from 'react';
+import { useCallback, useMemo, useRef, useState } from 'react';
 import { Toast } from 'primereact/toast';
 import { parseMapUserSettings } from '@/hooks/Mapper/components/helpers';
 import { saveTextFile } from '@/hooks/Mapper/utils/saveToFile.ts';
@@ -7,14 +7,18 @@ import { SplitButton } from 'primereact/splitbutton';
 import { loadTextFile } from '@/hooks/Mapper/utils';
 import { applyMigrations } from '@/hooks/Mapper/mapRootProvider/migrations';
 import { createDefaultStoredSettings } from '@/hooks/Mapper/mapRootProvider/helpers/createDefaultStoredSettings.ts';
+import { OutCommand } from '@/hooks/Mapper/types';
+import { WdButton } from '@/hooks/Mapper/components/ui-kit';
 
 export const ImportExport = () => {
   const {
     storedSettings: { getSettingsForExport, applySettings },
     data: { map_slug },
+    outCommand,
   } = useMapRootState();
 
   const toast = useRef<Toast | null>(null);
+  const [mapDataBusy, setMapDataBusy] = useState(false);
 
   const handleImportFromClipboard = useCallback(async () => {
     const text = await navigator.clipboard.readText();
@@ -143,6 +147,90 @@ export const ImportExport = () => {
     }
   }, [getSettingsForExport, map_slug]);
 
+  const handleExportMapData = useCallback(async () => {
+    setMapDataBusy(true);
+
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const res: any = await outCommand({ type: OutCommand.exportMapData, data: null });
+
+      if (!res?.data) {
+        throw new Error(res?.error ?? 'Empty response');
+      }
+
+      const { systems = [], connections = [], signatures = [] } = res.data;
+
+      saveTextFile(`map_${map_slug}_${new Date().toISOString().slice(0, 10)}.json`, JSON.stringify(res.data, null, 2));
+
+      toast.current?.show({
+        severity: 'success',
+        summary: 'Export map',
+        detail: `Saved ${systems.length} systems, ${connections.length} connections, ${signatures.length} signatures.`,
+        life: 4000,
+      });
+    } catch (error) {
+      console.error('Export map data Error: ', error);
+
+      toast.current?.show({
+        severity: 'error',
+        summary: 'Error',
+        detail: 'Some error occurred on exporting map data, check console log.',
+        life: 3000,
+      });
+    } finally {
+      setMapDataBusy(false);
+    }
+  }, [outCommand, map_slug]);
+
+  const handleImportMapData = useCallback(async () => {
+    let parsed;
+
+    try {
+      parsed = JSON.parse(await loadTextFile());
+    } catch (error) {
+      console.error('Import map data Error: ', error);
+
+      toast.current?.show({
+        severity: 'error',
+        summary: 'Error',
+        detail: 'Selected file is not a valid map export.',
+        life: 3000,
+      });
+      return;
+    }
+
+    setMapDataBusy(true);
+
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const res: any = await outCommand({ type: OutCommand.importMapData, data: { data: parsed } });
+
+      if (!res?.result) {
+        throw new Error(res?.error ?? 'Empty response');
+      }
+
+      const { systems, connections, signatures } = res.result;
+
+      toast.current?.show({
+        severity: 'success',
+        summary: 'Import map',
+        detail: `Added ${systems} systems, ${connections} connections, ${signatures} signatures.`,
+        life: 4000,
+      });
+    } catch (error) {
+      console.error('Import map data Error: ', error);
+
+      toast.current?.show({
+        severity: 'error',
+        summary: 'Error',
+        detail: 'Some error occurred on importing map data, check console log.',
+        life: 3000,
+      });
+    } finally {
+      setMapDataBusy(false);
+    }
+  }, [outCommand]);
+
   const importItems = useMemo(
     () => [
       {
@@ -198,6 +286,35 @@ export const ImportExport = () => {
         </div>
 
         <span className="text-stone-500 text-[12px]">*Will save map settings to clipboard.</span>
+      </div>
+
+      <div className="border-b-2 border-dotted border-stone-700/50 h-px" />
+
+      <div className="flex flex-col gap-1">
+        <div className="flex gap-2">
+          <WdButton
+            onClick={handleExportMapData}
+            icon="pi pi-cloud-download"
+            size="small"
+            label="Export map"
+            className="py-[4px]"
+            disabled={mapDataBusy}
+          />
+          <WdButton
+            onClick={handleImportMapData}
+            icon="pi pi-cloud-upload"
+            size="small"
+            severity="warning"
+            label="Import map"
+            className="py-[4px]"
+            disabled={mapDataBusy}
+          />
+        </div>
+
+        <span className="text-stone-500 text-[12px]">
+          *Map contents - systems, connections and signatures - as a file. Import adds what is missing, systems already
+          on the map are left untouched.
+        </span>
       </div>
 
       <Toast ref={toast} />
--- a/assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/components/ImportExport.tsx
+++ b/assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/components/ImportExport.tsx
@@ -9,6 +9,16 @@ import { applyMigrations } from '@/hooks/Mapper/mapRootProvider/migrations';
 import { createDefaultStoredSettings } from '@/hooks/Mapper/mapRootProvider/helpers/createDefaultStoredSettings.ts';
 import { OutCommand } from '@/hooks/Mapper/types';
 import { WdButton } from '@/hooks/Mapper/components/ui-kit';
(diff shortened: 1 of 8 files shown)


lib/wanderer_app/map/operations/transfer.ex
```diff
@@ -0,0 +1,298 @@
...
+             %{solar_system_id: solar_system_id, coordinates: position(system["position"])},
+             user_id,
+             character_id
+           ) do
+      # a system that was already on the map keeps whatever the map has for it
+      if MapSet.member?(existing_ids, solar_system_id) do
+        false
+      else
+        apply_system_attributes(map_id, solar_system_id, system)
+        true
```
guarzo, line 192: `existing_solar_system_ids/1` uses `read_all_by_map`, which has no `visible` filter — so it counts systems that were deleted from the map, and deletion here is a soft `visible: false` (`MapSystemRepo.remove_from_map/2`) with the row left behind.

That puts it out of step with the rest of the path. `Server.add_system` starts with `check_location/2` → `find_system_by_location/2`, which only returns a system when `visible: true`, so for a deleted one it says "not present" and goes ahead and adds it. Which is right — as far as anyone using the map is concerned that system isn't on it, and an import should bring it back like any other missing system.

The problem is what happens after: line 182 finds it in `existing_ids`, so `apply_system_attributes` is skipped and the function returns `false`. The system reappears with its labels, tags and temporary name cleared by the normal add path, but none of the imported `custom_name` / `description` / `labels` / `status` / `tag` / `temporary_name` / `locked` get applied on top — and it isn't counted, so the toast doesn't mention it either. Every genuinely-new system gets its attributes; this one gets nothing.

Switching `existing_solar_system_ids/1` to `MapSystemRepo.get_visible_by_map/1` lines it up with what `check_location` considers present, and then it's treated as new all the way through. It'd also match `export/2`, which already uses `get_visible_by_map` — worth having the two ends of the round trip agree on what's on a map.

lib/wanderer_app_web/controllers/map_api_controller.ex
```diff
@@ -1327,6 +1327,72 @@ defmodule WandererAppWeb.MapAPIController do
...
+      {:error, reason} ->
+        Logger.error("Map import failed: #{inspect(reason)}")
+
+        conn
+        |> put_status(:internal_server_error)
+        |> json(%{error: "Failed to import map"})
+    end
+  end
+
+  def import_map(conn, _params) do
```
guarzo, line 1395: I don't think this clause can be reached. `AssignMapOwner` always assigns both keys — when `get_owner_character_id/1` doesn't return an owner it does `assign(:owner_character_id, nil) |> assign(:owner_user_id, nil)` rather than leaving them off. So the map pattern on line 1357 still matches, `char_id` and `user_id` are just `nil`, and the import runs with a nil user and character instead of returning the 422.

`Transfer.import/5` is specced `binary(), map(), binary(), binary() | nil` too, so nil `user_id` is outside what it says it takes.

A guard on the first clause (`when not is_nil(char_id) and not is_nil(user_id)`) would make the fallback do what you meant. Worth a quick sanity check on whether the owner can genuinely be unresolvable for a map that has a working API key — if it can't, then the fallback is just belt-and-braces and a guard is still the cheapest way to make it honest.

lib/wanderer_app/map/operations/transfer.ex
```diff
@@ -0,0 +1,298 @@
...
+            |> Map.merge(%{"source" => to_string(source), "target" => to_string(target)})
+          ]
+        else
+          _ -> []
+        end
+      end)
+
+    Server.paste_connections(map_id, paste_payload, user_id, character_id)
+
+    length(paste_payload)
```
guarzo: The three numbers in the stats don't all mean the same thing, which shows up in the toast.

`imported_systems` counts systems that were genuinely new — that's the useful one. But `import_connections/4` returns `length(paste_payload)` here, which is every connection in the document with parseable endpoints, whether or not `paste_connections` did anything with it. And `create_signatures/2` below counts `{:ok, _}` from `MapSystemSignature.create`, which is an upsert on `uniq_system_eve_id` — so an already-present signature returns `{:ok, _}` and counts as imported.

Re-importing the same file into the same map therefore reports something like "Added 0 systems, 12 connections, 30 signatures" while having changed nothing. Since the pitch for this is that it's safe to re-run against a live chain, the counts are the main feedback people get about whether that's true.

(The upsert behaviour itself is right, and worth saying out loud: `upsert_fields` is only `[:system_id, :eve_id]`, so a re-import doesn't clobber names or descriptions someone has edited on the target. That's a good outcome — it's just not what "Added 30 signatures" tells them.)

lib/wanderer_app/map/operations/transfer.ex
```diff
@@ -0,0 +1,298 @@
...
+          []
+      end
+    end)
+  end
+
+  defp export_signature(signature, solar_system_id) do
+    %{
+      "solar_system_id" => solar_system_id,
+      "eve_id" => signature.eve_id,
+      "character_eve_id" => signature.character_eve_id,
```
guarzo: `character_eve_id` rides along in the document and gets written straight back on import.

Sits a bit oddly against the description, which says character-referencing data is deliberately left out because those characters don't exist on the importing side — that's exactly what this is. On a cross-deployment import the signature ends up attributed to a character id nobody on that instance has.

Not a crash risk (it's a plain `:string`, no FK), so this is really a question: is it deliberate that "who scanned this" survives the transfer? If not, dropping it from both `export_signature/2` and the `Map.take` in `create_signatures/2` would make the document match what the docs say it contains.

lib/wanderer_app/map/operations/transfer.ex
```diff
@@ -0,0 +1,298 @@
...
+  """
+
+  require Logger
+
+  alias WandererApp.Api.MapSystemSignature
+  alias WandererApp.Map.Server
+
+  @export_version 1
+
+  @type stats :: %{systems: non_neg_integer(), connections: non_neg_integer(), signatures: non_neg_integer()}
```
guarzo: `mix format` wants a few things — three spots across two files:

- this line, the `@type stats` breaks over four lines
- line 226, the `Map.take([...])` list
- `map_api_controller.ex:1339`, the `Transfer.export(map_id, include_signatures: ...)` call wants to go multiline (interestingly it's already written that way in the LiveView handler)

`mix format` sorts all three. (`router.ex` also fails `--check-formatted`, but on a double blank line at 428 that's already on `main` — not yours.)

lib/wanderer_app_web/live/map/event_handlers/map_core_event_handler.ex
```diff
@@ -209,6 +209,65 @@ defmodule WandererAppWeb.MapCoreEventHandler do
     {:noreply, socket}
   end
 
+  def handle_ui_event(
+        "export_map_data",
```
guarzo, line 215: Two small things about this pair of handlers:

`import_map_data` is gated on `user_permissions: %{add_system: true}` and `export_map_data` has no permission clause at all. Probably fine given anyone who can send this already sees the systems and signatures in the UI, but export is the one call that hands them the whole map as a file in one go, so it seems worth being a deliberate decision rather than an omission.

Also, both take `include_signatures` from params and the frontend never sends it — `handleExportMapData` posts `data: null`. So the option is supported at three layers and reachable from none of the UI. A checkbox next to the two buttons would be the obvious home for it, or it could come out of the LiveView path and stay an API-only thing.

assets/js/hooks/Mapper/components/mapRootContent/components/MapSettings/components/ImportExport.tsx
```diff
@@ -200,6 +288,35 @@ export const ImportExport = () => {
...
+        <div className="flex gap-2">
+          <WdButton
+            onClick={handleExportMapData}
+            icon="pi pi-cloud-download"
+            size="small"
+            label="Export map"
+            className="py-[4px]"
+            disabled={mapDataBusy}
+          />
+          <WdButton
```
guarzo, line 341: Importing writes to a live map as soon as the file picker closes — no confirmation, and no undo once systems and connections start landing. `severity="warning"` on the button is the only signal.

Given the map is shared, and the person importing may not have made the file, a confirm step ("this will add N systems and M connections to <map>, continue?") would be worth it. You've already got the document parsed client-side at that point, so the counts are free.

Somewhere in the same area — the export/import round trip through `handle_ui_event` is fully synchronous, and export runs one signature query per system. For a chain with a couple of hundred systems that's a few hundred queries with the LiveView blocked, and import does the writes and broadcasts inline too. Have you tried it against a big map? If it's fine at realistic sizes then no problem, I just couldn't tell from reading it where the ceiling is.

guarzo (commented):
Nice to see this on its own — it reads much better without the rest of #632 around it, and the moduledoc explaining how transfer differs from `Duplication` is the kind of thing I wish more modules had.

Two bits I got stuck on while reading. `existing_ids` is built from every row on the map rather than the visible ones, so a system somebody deleted counts as "already there" even though the map server doesn't think it is — it gets re-added but arrives with none of the imported attributes on it. And the 422 clause on `import_map` looks like it can't ever fire, which matters because it's the thing standing between a map with no resolvable owner and a nil `user_id`. Both inline, and I might be misreading either one.

Rest is smaller stuff — the counts in the toast, a `mix format` run, a couple of questions.

Two asks that don't have a natural line to hang off: `Duplication` has four test files (`test/unit/map_duplication_*.exs` plus an integration one) and this doesn't have any. The round trip is a good shape for one — export a map, import into an empty one, assert the counts and that a second import is a no-op. The deleted-system case below would fall out of that naturally (delete one on the target, re-import, check its labels came back). And since it adds two buttons to the settings dialog, a screenshot of the Import/Export tab would be great — you did that on #632 so I assume it just didn't come across in the split.

z4nr34l:
Thanks — this was a good read, and two of these were real. All addressed in `fa25812`.

**Deleted systems.** You were right, and it was worse than the attributes going missing: `read_all_by_map` disagreed with `check_location`, which goes by `visible`. Now on `get_visible_by_map`, so both ends of the round trip and the add path agree on what is on a map. There is a test for exactly your case — delete a system on the target, re-import, assert its attributes came back.

**The 422.** Also right, `AssignMapOwner` assigns `nil` rather than omitting the keys, so the map pattern always matched. Added the guard you suggested. I could not construct a map with a working API key and an unresolvable owner, so as far as I can tell it is belt-and-braces — but the spec says `binary()` for `user_id`, so it should be honest either way.

**The counts.** Fixed, and both were wrong in the way you describe. Connections now count only pairs the map did not already have (normalised, since a document can name the same pair from either end), and signatures diff against `by_system_id_and_eve_ids` before the upsert. A second import of the same file now reports zeroes, and the toast says "Everything in that file was already on the map" rather than a row of zeroes. That is the second test.

**`character_eve_id`.** Not deliberate, and interesting to dig into: it is `allow_nil? false` on the resource, so a document without it cannot be imported at all. Dropping it would have broken import outright. Signatures are now attributed to the character running the import, which is what the moduledoc claims and does not carry an id from another instance.

**While writing the test I found signatures were never exported at all.** `by_system_id_all(%{system_id: ...})` passes a map where the code interface takes a positional argument, so every call returned `{:error, %Ash.Error.Query.InvalidArgument{}}` — swallowed by the `{:error, _} -> []` clause below it, so exports quietly carried zero signatures. Confirmed by putting the old call back and watching `assert length(document["signatures"]) == 1` fail. It is fixed by the thing that also answers your performance question: one bulk read (`by_system_ids`) for the whole map instead of a query per system, so export is now one signature query regardless of map size.

Import still does its writes and broadcasts inline; I have not driven a couple-hundred-system chain through it, so I would not claim a ceiling I have not measured. If you want that bounded before this goes in, say the word.

**Permissions and `include_signatures`.** Export now asks for `view_system: true` explicitly rather than relying on the caller having got that far, with a matching fallback clause. `include_signatures` is now a checkbox next to the two buttons and is sent on both calls, so it is reachable rather than supported-at-three-layers-and-used-by-none.

**Confirmation before import.** Added — the document is parsed client-side first, so the dialog names the counts and the target map, and says it cannot be undone.

**`mix format`.** Run on all three spots. Left `router.ex` alone since that one is already on `main`.

**Tests.** `test/integration/map/map_transfer_test.exs`: round trip into an empty map (attributes, connection, signature), a second import as a no-op, a deleted system coming back with its attributes, and the two rejection paths. Four tests, passing.

Screenshots of the tab and the confirm dialog to follow shortly.

z4nr34l:
image
the confirm step: "This adds up to 14 systems, 17 connections and 42 signatures to this map, for everyone on the map. Systems already there are left alone, and an import cannot be undone."

wanderer-import-export-tab
the Import/Export tab with Export map / Import map and the new Include signatures checkbox

PR #657, opened 2026-08-11
fix(tracker): random loss of tracking, only fixed by relogging in EVE

### LINKED ISSUE #620: Loss of Tracking
Seemingly random loss of Tracking. Still shown on own view, not visible for other users. Location does not change when splashing on own view.
Fixable only by Relogging in EVE.
Removing and adding tracking from some or all characters does not fix this.

guarzo:
Tracking stops at random: the character stays visible on your own view but not to other users, and the position no longer updates when you jump. Removing and re-adding tracking does not help, only relogging in EVE does.

Fixes #620.

@DmitryPopov @DanSylvest

### DIFF
--- a/lib/wanderer_app/character/tracker.ex
+++ b/lib/wanderer_app/character/tracker.ex
@@ -1004,9 +1004,14 @@ defmodule WandererApp.Character.Tracker do
        ),
        do: state
 
+  # Matches on the state's active_maps rather than a `track_location` key in
+  # track_settings. No caller passes that key -- every call site sends
+  # %{map_id: _, track: true | false} -- so the previous clause never matched
+  # and this function could not turn location tracking back on once
+  # maybe_stop_tracking/2 had cleared it.
   defp maybe_start_location_tracking(
-         state,
-         %{track_location: true} = _track_settings
+         %{active_maps: [_ | _]} = state,
+         _track_settings
        ),
        do: %{state | track_location: true}
 
@@ -1017,8 +1022,8 @@ defmodule WandererApp.Character.Tracker do
        do: state
 
   defp maybe_start_ship_tracking(
-         state,
-         %{track_ship: true} = _track_settings
+         %{active_maps: [_ | _]} = state,
+         _track_settings
        ),
        do: %{state | track_ship: true}
 
--- a/test/unit/character/tracker_update_settings_test.exs
+++ b/test/unit/character/tracker_update_settings_test.exs
@@ -1,0 +1,131 @@
+defmodule WandererApp.Character.TrackerUpdateSettingsTest do
+  @moduledoc """
+  Regression tests for `WandererApp.Character.Tracker.update_settings/2`.
+
+  `maybe_start_location_tracking/2` used to match `%{track_location: true}` in
+  `track_settings`. No caller passes that key, so it never matched and the
+  function could not turn location tracking on. That was harmless while nothing
+  turned the flag off; once `maybe_stop_tracking/2` began clearing it, the flag
+  could get stuck off. The trigger is not a user action: when a browser's
+  presence lapses past the grace period the character is untracked server-side,
+  `active_maps` empties, and both flags are cleared. The character is still
+  online in EVE throughout, and `update_online/1` only writes the flag on an
+  online-status transition — which a character who stays logged in never
+  produces. So reconnecting could not restore it.
+  """
+
+  use ExUnit.Case, async: false
+
+  alias WandererApp.Character.Tracker
+
+  setup do
+    character_id = "test-char-#{System.unique_integer([:positive])}"
+    map_id = "test-map-#{System.unique_integer([:positive])}"
+
+    on_exit(fn -> Cachex.del(:character_state_cache, character_id) end)
+
+    %{character_id: character_id, map_id: map_id}
+  end
+
+  defp seed_state(character_id, overrides) do
+    state =
+      Tracker.new(%{character_id: character_id})
+      |> Map.merge(overrides)
+
+    Cachex.put(:character_state_cache, character_id, state)
+    state
+  end
+
+  describe "update_settings/2 when map tracking starts" do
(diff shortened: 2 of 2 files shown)


guarzo:
## Proposal

Reported in #620: tracking stops at unpredictable times, position no longer updates when splashing, and — the diagnostic detail — **toggling tracking off and on does not fix it, only relogging in EVE does.** That asymmetry is what this patch explains and repairs.

A character who is online in EVE stops moving on the map. They stay logged in, stay present, stay green — their position simply never updates again. Nothing errors and nothing logs.

**The user does nothing to trigger this.** They never leave a map or untrack a character. The map removal happens server-side, driven by the *browser* websocket: `PresenceGracePeriodManager` runs a 15-minute grace period (`lib/wanderer_app_web/presence_grace_period_manager.ex:29` — the one web-layer citation here; everything else below is under `lib/wanderer_app/`), and when it expires the character is dropped from `map_<id>:presence_character_ids` (`:384`). A closed tab, a slept laptop, or a network blip is enough. `untrack_characters/2` then fires — `map_server_characters_impl.ex:80` logs it as *"characters no longer in presence_character_ids (grace period expired or user disconnected)"* and emits `reason: :presence_expired` — which calls `update_track_settings(%{map_id:, track: false})`, empties `active_maps`, and lets `maybe_stop_tracking/2` clear both flags.

The character is still in EVE, still flying, throughout. When the browser reconnects, tracking is re-enabled with `%{track: true}` — and that is the path that cannot restore the flag.

## The defect

```elixir
defp maybe_start_location_tracking(state, %{track_location: true} = _track_settings),
  do: %{state | track_location: true}

defp maybe_start_location_tracking(state, _track_settings), do: state
```

The clause matches `track_location: true` in **`track_settings`** — the caller-supplied map. No caller passes that key. All five call sites send `%{map_id: _, track: true | false}`:

| call site (under `lib/`) | settings |
|---|---|
| `wanderer_app/character/tracking_utils.ex:398` | `%{map_id:, track: true}` |
| `wanderer_app/map/server/map_server_characters_impl.ex:110` | `%{map_id:, track: false}` |
| `wanderer_app/map/server/map_server_characters_impl.ex:1077` | `%{map_id:, track: true}` |
| `wanderer_app/map/server/map_server_characters_impl.ex:1103` | `%{map_id:, track: true}` |
| `wanderer_app/character/tracker_manager_impl.ex:371` | `%{map_id:, track: false}` |

So the clause never matches, the catch-all returns `state` unchanged, and this function cannot turn location tracking on. `maybe_start_ship_tracking/2` has the identical defect.

## Why that strands a character

`track_location` has exactly three writers, all in `lib/wanderer_app/character/tracker.ex` (note: unrelated to the `:1103` in the call-site table above, which is a different file):

| line | writes | when |
|---|---|---|
| `tracker.ex:17` | `false` | struct default |
| `tracker.ex:191` | `online.online` | only inside `if online.online != is_online` — an online-status **transition** |
| `tracker.ex:1103` | `false` | `maybe_stop_tracking/2`, once `active_maps` is empty — reached via presence expiry, not a user action |

From `(is_online: true, track_location: false)` there is no way out. Line 17 is unreachable, line 1103 only re-clears, and line 191 needs a transition that a character who stays logged into EVE never produces. The dead clause was the only path back.

`update_location/1` requires `%{track_location: true, is_online: true}` and otherwise falls through to a catch-all, so every one-second tick is skipped silently. And because `track_online` is untouched by any of this, online polling keeps running: the character stays present, stays green, and every health signal reports fine around a position that never changes.

## Why #620's workaround is the only one that works

Toggling tracking off and on calls `update_settings/2`, which reaches `maybe_start_location_tracking/2` — the dead clause. It cannot restore the flag, so the character comes back still frozen. Relogging in EVE drives `is_online` `true → false → true`; each edge is a transition, so line 191 runs and restores `track_location` as a side effect. The reporter found the only working remedy and correctly ruled out the obvious one.

One detail from #620 I have not traced: the report notes the character stays visible on their own view while other users can no longer see them. The freeze itself is fully explained above; the disappearance is most likely the `broadcast!(map_id, :untrack_character, character_id)` that fires when the untrack lands — the same event that clears the flag. That is consistent but inferred, not verified.

## Why it hasn't surfaced sooner

The clause has been dead since the file was written, but it was harmless while nothing cleared the flag. The struct default was `track_location: true` and `maybe_stop_tracking/2` ended with a bare `state`. A flag that starts true and is never cleared needs no setter.

`f761ba97` changed both — defaults to `false`, and `maybe_stop_tracking/2` gained `%{state | track_location: false, track_ship: false}`. That commit didn't introduce the defect; it removed the conditions that were masking it.

## The fix

Match on the state's `active_maps` — the thing that is actually true when polling should be on:

```elixir
defp maybe_start_location_tracking(%{active_maps: [_ | _]} = state, _track_settings),
  do: %{state | track_location: true}
```

Same for ship tracking. Both flags are intent flags, not liveness flags — `update_location/1` independently requires `is_online: true` — so setting them for an offline character is safe and avoids depending on a transition that may never arrive.

`maybe_start_online_tracking/2` has the same never-matching clause and is deliberately left alone: `track_online` defaults to `true` at `tracker.ex:16` and nothing anywhere clears it, so its setter has nothing to restore. Only the two flags `maybe_stop_tracking/2` clears can get stuck, so only those two are changed here.

## Tests

`test/unit/character/tracker_update_settings_test.exs`, four cases: restore while online (the stuck state), enable while offline, idempotence on repeat track, and clearing once no maps remain. Verified to fail without the fix — 2 of 4 fail, including the restore case.

## Verification

`mix test` on `main` before and after gives the **same 12 failures** (6 `MapDuplicationAPIControllerSuccessTest`, 6 `AuthControllerTest`), all pre-existing and unrelated to tracking. `mix compile --warnings-as-errors` clean. Both changed files pass `mix format --check-formatted` (several unrelated files on `main` do not).

Fixes #620

PR #664, opened 2026-09-09
fix: make map webhooks usable

guarzo:
Enabling webhooks returned “Webhook not found,” and creating a webhook failed with “Invalid attribute: secret.” Webhook toggling, creation, and secret rotation now work correctly.

@DmitryPopov @DanSylvest

### DIFF
--- a/lib/wanderer_app/api/map_webhook_subscription.ex
+++ b/lib/wanderer_app/api/map_webhook_subscription.ex
@@ -138,7 +138,7 @@ defmodule WandererApp.Api.MapWebhookSubscription do
       # Generate secret on creation
       change fn changeset, _context ->
         secret = generate_webhook_secret()
-        Ash.Changeset.force_change_attribute(changeset, :secret, secret)
+        AshCloak.encrypt_and_set(changeset, :secret, secret)
       end
 
       # Invalidate cache when subscription is created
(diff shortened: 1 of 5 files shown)


guarzo:
## Proposal

The webhook toggle route was declared after `/webhooks/:id`, so Phoenix treated `toggle` as a webhook ID and sent the request to the update action. The toggle action also expected `map_id` even though its route supplies `map_identifier`.

Webhook secrets are encrypted with AshCloak, which replaces the writable `secret` attribute with an encrypted backing attribute. Creation and rotation attempted to write directly to `secret`, causing `NoSuchAttribute`.

This change:

- declares the static toggle route before webhook resource routes
- makes `map_identifier` authoritative while retaining `map_id` compatibility for direct callers
- prevents body parameters from overriding the authenticated path map
- generates and rotates secrets through `AshCloak.encrypt_and_set/3`
- adds regression coverage for routing, identifier precedence, creation, encryption, and rotation

No database migration or Ash resource snapshot change is needed because the resource schema is unchanged.

Verification:

- focused webhook and controller suite: 28 tests, 0 failures
- compile with warnings as errors: passed
- full suite at seed 0: 936 tests, 13 failures
- clean `main` at seed 0: 931 tests, the same 13 baseline failures
- `git diff --check`: passed

PR #665, opened 2026-09-10
Jump Planner - Added sidebar widget on Wanderer side which allow to setup Jump settings. 

DanSylvest:
Back to the game and discovered that old instrument for opening Jump Planner looks and works wierd. Just want to make it little bit easier for using

image
image
image
image
image

Please note: this is not a complete in-app jump planner yet. It helps you prepare the route and settings, but the actual jump route still needs to be opened and calculated in DOTLAN.

**What’s included:**
• Open the setup tool from the right sidebar or Map Menu
• Use **Jump From** and **Jump To** from system context menus
• Access the same actions from the Routes widget
• Search and select origin and destination systems
• Use High-Sec, Low-Sec, or Null-Sec systems as the origin
• Select only Low-Sec or Null-Sec systems as the destination
• Choose a specific jump-capable ship from categorized ship groups
• Configure Jump Drive Calibration, Jump Fuel Conservation, and Jump Freighter skill levels
• Configure station preference and incursion avoidance
• Save ship and planner settings locally
• Open the configured route in DOTLAN to calculate and review it

The relevant field is focused automatically when opening the tool from a system, making DOTLAN route setup faster and more convenient.


assets/js/hooks/Mapper/components/mapRootContent/components/JumpPlanner/JumpPlanner.tsx
```diff
@@ -0,0 +1,375 @@
...
+  if (!system || !isPossibleSpace(allowedSystemClasses, system.system_class)) {
+    return null;
+  }
+
+  return toSearchSystemItem(system);
+};
+
+const getJumpPlannerUrl = (settings: typeof DEFAULT_JUMP_PLANNER_SETTINGS, from: string, destination: string) => {
+  const { shipType, jumpDriveCalibration, jumpFuelConservation, jumpFreighter } = settings;
+  let ship = `${encodeURIComponent(shipType)},${jumpDriveCalibration}${jumpFuelConservation}${jumpFreighter}`;
```
guarzo: Could we normalize spaces to underscores before URL-encoding the ship and system names? DOTLAN doesn't accept `%20` in these positions.

I checked the generated URLs against DOTLAN: selecting `Revelation Navy Issue` redirects to the blank planner, while routes containing `New Caldari` or `Old Man Star` lose their endpoints. The corresponding underscore versions work.

Applying `encodeURIComponent(name.replaceAll(' ', '_'))` to each name should address this. A regression test with a multiword ship and system would help cover it.

assets/js/hooks/Mapper/components/mapRootContent/components/JumpPlanner/JumpPlanner.tsx
```diff
@@ -0,0 +1,375 @@
...
+      field="label"
+      placeholder={placeholder}
+      emptyMessage="Not found any system..."
+      showEmptyMessage
+      minLength={SYSTEM_SEARCH_MIN_LENGTH}
+      scrollHeight="300px"
+      autoComplete="off"
+      forceSelection
+      multiple
+      selectionLimit={1}
```
guarzo: Could we keep the selected system clearable from the keyboard? In PrimeReact 10.6.5, `multiple` combined with `selectionLimit={1}` makes the input read-only and removes its keyboard handler once a value is selected. Backspace no longer clears it, and the remaining removal icon isn't keyboard-focusable.

This affects both manually selected and context-menu-prefilled systems. Removing `selectionLimit` while retaining the existing last-item-only `onChange` handling may be enough; otherwise, an accessible clear button would solve it.

assets/js/hooks/Mapper/components/mapRootContent/components/JumpPlanner/JumpPlanner.tsx
```diff
@@ -0,0 +1,375 @@
...
+
+  return toSearchSystemItem(system);
+};
+
+const getJumpPlannerUrl = (settings: typeof DEFAULT_JUMP_PLANNER_SETTINGS, from: string, destination: string) => {
+  const { shipType, jumpDriveCalibration, jumpFuelConservation, jumpFreighter } = settings;
+  let ship = `${encodeURIComponent(shipType)},${jumpDriveCalibration}${jumpFuelConservation}${jumpFreighter}`;
+  if (settings.preferStationSystems) {
+    ship += ',S';
+  }
```
guarzo: There is an integration edge case here: omitting `,S` doesn't necessarily disable station preference on DOTLAN.

I requested the same route URL with and without DOTLAN's `jp_station=1` cookie. DOTLAN enabled “Prefer Station Systems” with the cookie and disabled it without, even though neither URL contained `,S`. A user who previously saved that preference on DOTLAN can therefore uncheck it here and still get a station-preferred route.

Could we check whether DOTLAN supports an explicit override? If it doesn't, we should explain that saved DOTLAN preferences can override this unchecked option.

assets/js/hooks/Mapper/components/mapRootContent/components/RightBar/RightBar.tsx
```diff
@@ -99,12 +105,24 @@ export const RightBar = ({
                   type="button"
                   onClick={onShowWormholesReference}
                 >
-                  <i className="pi pi-bullseye"></i>
+                  <i className="pi pi-book"></i>
+                </button>
+              </WdTooltipWrapper>
+
+              <WdTooltipWrapper content="Jump Planner" position={TooltipPosition.left}>
```
guarzo, line 112: Could we move the Jump Planner button outside the `canTrackCharacters` block? The planner doesn't require character tracking, and its Map Menu entry is already available regardless of that permission.

Currently, users without tracking permission can open it from Map Menu but don't get the right-sidebar shortcut. Moving this button outside the tracking-only section would make the two entry points consistent.

assets/js/hooks/Mapper/components/mapRootContent/components/JumpPlanner/JumpPlanner.tsx
```diff
@@ -0,0 +1,375 @@
...
+        setSuggestions([]);
+      }
+    },
+    [allowedSystemClasses, outCommand],
+  );
+
+  return (
+    <AutoComplete
+      ref={inputRef}
+      id={id}
```
guarzo: Could this use `inputId={id}` rather than `id={id}`? PrimeReact applies `id` to the AutoComplete wrapper, so the surrounding label's `htmlFor` points to a `<span>` instead of the input.

I checked the rendered markup: the native input has no ID or accessible label, and `label.control` is `null`. Using `inputId` would connect the existing “From” and “Destination” labels to their inputs.

assets/js/hooks/Mapper/components/mapRootContent/components/RightBar/RightBar.tsx
DanSylvest: Reasonable but stay it as is.

assets/js/hooks/Mapper/components/mapRootContent/components/JumpPlanner/JumpPlanner.tsx
DanSylvest: Looks like this dotlan issue. And yep. Add hint about it

assets/js/hooks/Mapper/components/mapRootContent/components/JumpPlanner/JumpPlanner.tsx
DanSylvest: done

assets/js/hooks/Mapper/components/mapRootContent/components/JumpPlanner/JumpPlanner.tsx
DanSylvest: Fixed

